PR #4, opened 2025-05-27
Suppress artifacting in fireplace

asundqui:
Set additive opacity to 0 in fireplace so we don't highlight artifacts. Shouldn't really be adding to opacity to the highlight, not a big deal usually but this makes it so only the RGB is changed by the fireplace SDF.

After this change:
image

Before this change:
image


dmarcos:
Thanks so much

PR #5, opened 2025-05-27
Fix CORS image loading issue in procedural-splats.

asundqui:
Small tweak to make the image splats show up in examples/procedural-splats.


dmarcos:
Thanks!

PR #15, opened 2025-05-28
minor README updates

winnie1994:
Suggested README fixes given the current repo state! Feel free to veto any. 
(Asides from typos, these are mostly minor things that might help any beta testers coming in.)


dmarcos:
Thanks so much for the patience!

PR #16, opened 2025-05-28
Use `vite-plugin-dts`

dylanhu7:
Fixes TypeScript `.d.ts`. files. Have tried with option that rolls up types into a single `forge.d.ts` file but unfortunately this was having issues, perhaps related to various Vite config options we are using.

What's currently on `main` is the output of `tsc` itself, but for some reason this does not `export` any modules, so it does not work, at least not through the LSP.

This is the only solution that works that I have been able to find, but there may be better solutions we can explore in the future. But this will get us into a working state wrt TypeScript.


dmarcos:
Can we put the types in a `dist/types` directory or something like that to keep it clean?

dylanhu7:
> Can we put the types in a `dist/types` directory or something like that to keep it clean?

That's a good idea - done

dmarcos:
Good work! Thanks for the patience and flexibility

PR #21, opened 2025-05-29
Add repo link button

alex-mcneilly:
I added a button in the bottom right-hand corner of each example that links to the code in the repo for that example:

Screenshot 2025-05-29 at 10 30 12 AM

This is based on the repo button in the lower right-hand corner on the Three.js examples page:

Screenshot 2025-05-29 at 10 28 45 AM

If the styling, placement, or links should change, please let me know!

### DIFF
--- a/examples/vfx-dynamic-lighting/index.html
+++ b/examples/vfx-dynamic-lighting/index.html
@@ -42,7 +42,7 @@
     <label><input type="checkbox" id="debug" autocomplete="off"> Debug Light SDFs</label>
     <label><input type="checkbox" id="daynight" checked="checked" autocomplete="off"> Enable Lighting</label>
   </header>
-  
+
   <canvas id="canvas"></canvas>
   <script type="importmap">
   {


dmarcos:
Problem here is that it adds manual labor, maintaainace and makes each example more complicated and harder to copy and paste. I believe three.js renders the examples in an iframe and overlays a button on top without modifying the examples.

dmarcos:
Let me know when this is ready. Noticed the examples.json. Wonder if we can avoid and keep the original [index.html](https://github.com/forge-gfx/forge/blob/main/index.html). Probably can keep it simple for now. Do the iframe examples page for the docs and duplicate the list of examples for now. Can do something fancier later

alex-mcneilly:
Ok. It will be ready once I fix some styling things for the mobile version. Very close to finishing right now. Do you not want to use the `examples.json` anymore and not dynamically load them from a JSON file? Also, when you say to keep the original `index.html` and just put the iframe examples in the docs, does that mean I should put what I've written under the `docs` directory?

dmarcos:
To keep simple for now leave the old `index.html` and create an `examples.html` in the root directory. Would avoid an examples.json for now. Just duplicate the list manually. When deploying the site I'll copy `examples.html` to the right place.

Hopefully I'm not making it too confusing :)

alex-mcneilly:
Ok. Sounds good. I'll work on this. Thank you for clarifying!

alex-mcneilly:
@dmarcos @asundqui 
Ok. I've finished the examples page and put it in `examples.html`. It is no longer in `examples/index.html`, and I have reverted that to what it originally was. We are also no longer using `examples.json`.

This is what the desktop version of the `examples.html` page looks like:
Screenshot 2025-05-30 at 1 41 10 PM


This is what the mobile version of the `examples.html` page looks like:
Screenshot 2025-05-30 at 1 41 48 PM
Screenshot 2025-05-30 at 1 41 35 PM

dmarcos:
Looking good! left comments with couple of nits

alex-mcneilly:
Ok. I fixed the nits. It should be ready to merge unless I should change something else.

dmarcos:
Cool stuff! Awesome work. Gonna make the site so much nicer. Thanks

PR #22, opened 2025-05-29
Fix snow example

alex-mcneilly:
I've fixed the snow example to shift the snowbox, splat, and camera to the origin so the snow particles are less choppy at smaller fall velocity. I've adjusted the snowbox bounds and positions accordingly.


dmarcos:
Thanks

PR #23, opened 2025-05-30
wip pieces for integration test run

winnie1994:
WIP snippets for doing integration exercise with https://github.com/n0neye/A3D

Outstanding (potentially A3D side things?) things not caught in this PR that in A3D code:

1. dynamically importing forge, as forge import is triggering server side rendering errors in Next.js setup

Some unresolved issues
1. handling spz file loading in electron -- the electron file reader works for uncompressed formats, doesn't for `spz`. I'm assuming might just need to repro the gunzip setup in forge but not super familiar with it.
2. loading multiple splats in the A3D entity setup causes rendering artefacts. Guess it might be related to the `three.group.add(splat)/three.object3d().add(splat)` hierarchy thing but don't have an isolated example yet

Some needed features (for later)
1. depth map rendering

### DIFF
--- a/dist/forge.module.js
+++ b/dist/forge.module.js
@@ -8288,7 +8288,7 @@ function WorkerWrapper(options) {
   try {
     objURL = blob && (self.URL || self.webkitURL).createObjectURL(blob);
     if (!objURL) throw "";
-    const worker = new Worker(objURL, {
+    const worker = new window.Worker(objURL, {
       name: options == null ? void 0 : options.name
     });
     worker.addEventListener("error", () => {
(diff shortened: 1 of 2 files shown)


dmarcos:
Dist is all the source files compiled to JS + dependencies put together. PR should be against `src` files. I see though the code related to workers might come from a dependency. Before a fix though would be cool to understand why those `window.Worker` are needed in the A3D context

winnie1994:
Makes sense! I moved the object hierarchy change into `src` for now. 

A3D uses next.js so it's basically [this thread](https://github.com/vercel/next.js/issues/74621#issuecomment-2585760180) here. It seems like this might have been fixed on the next.js side 2 weeks ago but I did also try with `npm install next@latest` but that didn't help?

dmarcos:
Thanks. it looks like Workers issue is something Next.js should handle properly.

winnie1994:
Sounds good 😆 I just reverted all changes to `dist/` for now!

winnie1994:
pausing on this for now given side discussion! ideally would have a more proper solution that avoids traversing scene hierarchy every time an object gets added 👍

PR #32, opened 2025-06-03
Add stats.js for frame stats with toggle in GUI.

asundqui:
Added fps/frame time stats to examples/editor using stats.js (already in package.json).
Also added a GUI checkbox to toggle it on and off.


dmarcos:
thanks

PR #37, opened 2025-06-06
Improve rendering of highly-transparent splat scans

asundqui:
In `splatDefines.glsl` we define `MIN_ALPHA = 0.01`, which sets the minimal opacity of a splat before it is culled. This works fine for many splats, but for splat scans that are highly transparent, this means that splats that are below 1% opacity will be culled and never rendered.

Opacity is encoded in a Uint8, which means the lowest encodable value is actually 1/255 =0.00392, so splats with alpha values of 1/255 (0.0039) and 2/255 (0.0078) will simply be discarded. This PR lowers the MIN_ALPHA threshold to 0.5/255 (0.00196), which provides an additional buffer for low-opacity splats allowing them to be rounded up to 1/255.

Before the change:
min_alpha_0 01

After the change:
min_alpha_0 0019

Additionally, added the "sortRadial" option to `examples/editor` to allow users to toggle between Z-depth sorting (more common) and radial sorting (less common, but more stable under rotation). Also lowered Z-near clipping plane in the viewer to 0.01 from 0.1 so smaller scans are easier to see without culling.

### DIFF
--- a/examples/editor/index.html
+++ b/examples/editor/index.html
@@ -72,7 +72,7 @@
     import { getAssetFileURL } from "/examples/js/get-asset-url.js";
 
     const scene = new THREE.Scene();
-    const camera = new THREE.PerspectiveCamera(75, window.innerWidth / window.innerHeight, 0.1, 1000);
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks!

PR #45, opened 2025-06-11
Add `types` field to `package.json`

dylanhu7:
Adds a `types` field to `package.json` so that Node-style module resolution will pick up the types. This is necessary for projects that have `"moduleResolution": "node"` in their `tsconfig.json`.


dmarcos:
Thanks so much

PR #56, opened 2025-06-20
Add GLSL inline shader code injection example

asundqui:
Added an example called "GLSL Shaders" that shows how to write GLSL code inline in a SplatMesh modifier pipeline.

Has two examples: One is a direct GLSL code injection on the worldModifier to modulate color as a function of world x/y/z coordinate.

The second example injects a DynoBlock into objectModifier, and then GLSL dyno inside to show how to use other external variables as inputs, in this case a "time" uniform that is used to animate the rippling motion.

https://github.com/user-attachments/assets/778cbd4a-681b-4cbe-8f28-e867171f66ae

Thoughts feedback, is this a good starting point for people to start hacking GLSL code into splats? @dmarcos  does this example show a clean simple way to start writing GLSL code do you think?

### DIFF
--- a/examples/glsl/index.html
+++ b/examples/glsl/index.html
@@ -10,7 +10,7 @@
       margin: 0;
     }
     canvas {
-        touch-action: none;
+      touch-action: none;
     }
   </style>
 </head>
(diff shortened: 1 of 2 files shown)


dylanhu7 (approved):

dmarcos:
Thanks so much!

PR #58, opened 2025-06-20
Add Normal and Depth coloring modifiers and example

asundqui:
Created modifiers/ folder with normalColor and depthColor to start. Can be used by importing `modifiers` from Spark and applying those to SplatMeshes.

Added another examples/debug-color to show using both in an example.

The normal coloring is applied in "world space", with RGB corresponding to XYZ.

The depth coloring allows you to specify minimum and max Z depth, and logarithmically encodes depth from black to white on that scale (including boolean to reverse direction). Future PR may involve mapping these values over a color palette. @winnie1994  thanks for your example code, hopefully this depth coloring is good enough for people to use for depth conditioning etc?

https://github.com/user-attachments/assets/41a0a4ec-f0ef-4a1e-9e57-9d468d8074a3


61cygni:
Looks great!

PR #63, opened 2025-06-22
Add CodeMirror editor to examples/glsl

prasound24:
The idea is to live edit `.objectModifier` with an inline code editor:

codemirror

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,8 @@
     "spark-internal-rs": "file:rust/spark-internal-rs/pkg"
   },
   "dependencies": {
+    "codemirror": "6.0.2",
+    "codemirror-lang-glsl": "0.5.0",
     "fflate": "^0.8.2"
   },
   "keywords": ["3d", "three.js", "gsplats", "3dgs", "gaussian", "splats"]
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -58,6 +58,8 @@
     "spark-internal-rs": "file:rust/spark-internal-rs/pkg"
   },
   "dependencies": {
+    "codemirror": "6.0.2",
+    "codemirror-lang-glsl": "0.5.0",
```
dmarcos, line 62: This dependencies will be bundled with spark. Gotta go somewhere else. Can we use CDN?

dmarcos:
Thanks. It looks really cool. I would prefer keeping the example as simple as possible for educational purposes.

dmarcos:
It would be really cool to link this live editor demo from the [awesome repo](https://github.com/sparkjsdev/awesome). What do you think? Thanks so much for putting the time

soundform:
Do you want me to keep it in `examples/glsl` or move it to `sparkjsdev/awesome`?

dmarcos:
> Do you want me to keep it in `examples/glsl` or move it to `sparkjsdev/awesome`?

Yeah. You can submit a link to https://github.com/sparkjsdev/awesome/blob/main/README.md#demos

soundform:
Here is a PR for that repository: https://github.com/sparkjsdev/awesome/pull/2

PR #71, opened 2025-06-27
Use `getDrawingBufferSize` to determine renderer size

mrxz:
The `rendererSize` was retrieved using `WebGLRenderer.getSize(target)`. However, this method returns the size in _logical pixels_, whereas the actual size in _physical pixels_ is needed. Using `getDrawingBufferSize` resolves this issue and returns the correct value for High-DPI displays (when `pixelRatio != 1`) and during WebXR sessions (both with and without WebXR Layers)

The AVP specific workaround puzzles me a bit. Unless this was written against an old version of Three.js, the size should be kept in sync. So `getSize()` returning `1x1` would imply a `1x1` render target being created for WebXR, which would be entirely unusable. @dmarcos  Could you verify that rendering on an AVP works with these changes and that the workaround is no longer needed?

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -418,16 +418,7 @@ export class SparkRenderer extends THREE.Mesh {
       );
     } else {
       // Rendering to the canvas or WebXR
-      const renderSize = renderer.getSize(this.uniforms.renderSize.value);
-      if (renderSize.x === 1 && renderSize.y === 1) {
-        // WebXR mode on Apple Vision Pro returns 1x1 when presenting.
-        // Use a different means to figure out the render size.
-        const baseLayer = renderer.xr.getSession()?.renderState.baseLayer;
-        if (baseLayer) {
-          renderSize.x = baseLayer.framebufferWidth;
-          renderSize.y = baseLayer.framebufferHeight;
-        }
-      }
+      renderer.getDrawingBufferSize(this.uniforms.renderSize.value);
     }
 
     // Update uniforms from instance properties


dmarcos:
Thanks so much. I’ll try later today when I have my AVP available.

mrxz:
Here's an example that shows the difference it makes. Note that the effect is more pronounced in the image compared to when viewed on an actual high DPI display:
difference

Splat source: https://note.com/steam_studio/n/ne9736d94f162 (using the `cactus_splat3_30kSteps_142k_splats.compressed.ply` file)

asundqui:
@mrxz this is truly an amazing find! Thank you so much for digging in and really getting to the root of what was going on.

I tested your fix and yes can confirm it makes it sharper. I tested on Vision Pro and for me unfortunately that strange 1x1 size workaround was necessary to make it work. I tested Quest 3 and it works correctly with this workout. I've pushed the workaround to this PR.

@dmarcos I think we should merge this in, do you want to check my workaround on Vision Pro?

dmarcos:
Awesome work! Thanks so much. @mrxz workaround acceptable for now? I’ll verify on AVP before release

mrxz:
Thanks for confirming that the workaround for the AVP is actually still needed. Definitely a weird quirk of Apple's WebXR implementation. Ideally this will be fixed upstream, but in the meantime having this workaround is acceptable. I'm going to open an issue on the [immersive-web/webxr](https://github.com/immersive-web/webxr) repo as I'm not sure Apple is spec compliant here or not.

See: https://github.com/immersive-web/webxr/issues/1410

dmarcos:
Thanks so much!

PR #73, opened 2025-06-28
Added PC SOGS file format support

asundqui:
Implemented SOGS file format loading support matching this writer: https://github.com/playcanvas/sogs

The PlayCanvas SOGS file format begins with a `meta.json` root file that references multiple WEBP image files, each containing a layer of the properties, organized in a 2D grid to bring similar colors closer together. Although modern browsers support decoding WEBP into images, they do not allow accessing the underlying RGBA data directly. The images can be written to a canvas which can then be read back, but RGBA values are blended into the canvas by multiplying the RGB components with A, and cannot be inverted without losing precision.

To solve this I added the dependency @jsquash/webp which seems to be one of the best and cleanest implementations. Unfortunately it requires WASM for decoding, and I'm hoping this does not create bundling issues. I looked around for pure-JS implementations but did not find a suitable one.

Because the PC SOGS format requires multiple files to decode, it's not possible to fetch just one file and pass in `fileBytes` to initialize a `SplatMesh`. Passing in a `url` argument pointing to `meta.json` will trigger `SplatLoader` to try parsing it as a JSON first and check if the shape matches what's expected. If so, it will download the referenced files in `extraFiles` to be available for parsing.

Updated both viewer and editor to accept the files, changing the logic slightly to allow `SplatMesh` to fetch the URL data rather than fetching `fileBytes` and passing that in.

I validated the results with a number of scans, comparing the original PLY with the PlayCanvas SOGS converter and comparing it manually against the PLY.

### DIFF
--- a/examples/editor/index.html
+++ b/examples/editor/index.html
@@ -68,7 +68,7 @@
     import * as THREE from "three";
     import { OrbitControls } from "three/addons/controls/OrbitControls.js";
     import { GUI } from "lil-gui";
-    import { constructGrid, SparkControls, SparkRenderer, SplatMesh, textSplats, dyno, transcodeSpz } from "@sparkjsdev/spark";
+    import { constructGrid, SparkControls, SparkRenderer, SplatMesh, textSplats, dyno, transcodeSpz, isPcSogs } from "@sparkjsdev/spark";
     import { getAssetFileURL } from "/examples/js/get-asset-url.js";
 
     const scene = new THREE.Scene();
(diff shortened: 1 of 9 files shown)


dmarcos:
Thanks

mrxz:
> Although modern browsers support decoding WEBP into images, they do not allow accessing the underlying RGBA data directly. The images can be written to a canvas which can then be read back, but RGBA values are blended into the canvas by multiplying the RGB components with A, and cannot be inverted without losing precision.

@asundqui It's possible to get the raw RGBA data using a webgl2 context of the canvas instead of a 2d canvas.

asundqui:
> @asundqui It's possible to get the raw RGBA data using a webgl2 context of the canvas instead of a 2d canvas.

@mrxz 
Interesting, I didn't try that, but I think PlayCanvas' implementation does that...

I would love to remove the new dependency and only use web apis... Would you be interested in helping submit a follow-up PR to so this?

mrxz:
> Would you be interested in helping submit a follow-up PR to so this?

@asundqui See #90 for a PR that does a straightforward replacement. The mentioned downsides are mostly theoretical. I'd be surprised if any webp image used in SOGS files would actually show differences between browsers. And in case the device can't handle the image's dimensions as a texture, it most likely won't be capable of rendering gaussians splats in the first place.

PR #80, opened 2025-06-30
Use RenderTarget properties to reduce manual render state tracking

mrxz:
The `PackedSplats` and `Readback` classes kept track of various properties of the render state in order to restore them afterwards. The `WebGLRenderTarget` has properties for [`scissor/scissorTest`](https://threejs.org/docs/index.html?q=render#api/en/renderers/WebGLRenderTarget.scissor) which, when used, cause the `WebGLRenderer` to keep track of and restore them for us when switching render targets. This simplifies the code as there is less state to track and no longer a need to adjust the `pixelRatio` (as this doesn't apply to render targets).

I've also changed the `xrPresenting` to `xrEnabled`, since the `isPresenting` flag is considered readonly in Three.js ([WebXRManager.js#L100-L107](https://github.com/mrdoob/three.js/blob/dev/src/renderers/webxr/WebXRManager.js#L100-L107)). Temporarily disabling the WebXR code paths in the renderer should be done by setting `xr.enabled` to false, as can also be seen in the `Reflector` for example ([Reflector.js#L201](https://github.com/mrdoob/three.js/blob/dev/examples/jsm/objects/Reflector.js#L201)).

### DIFF
--- a/src/PackedSplats.ts
+++ b/src/PackedSplats.ts
@@ -364,6 +364,7 @@ export class PackedSplats {
     this.target.texture.format = THREE.RGBAIntegerFormat;
     this.target.texture.type = THREE.UnsignedIntType;
     this.target.texture.internalFormat = "RGBA32UI";
+    this.target.scissorTest = true;
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


dmarcos:
Thanks. Looks good to me. @asundqui good to go?

asundqui:
Probably good to go, but can I test on Quest 3 and AVP tonight to make sure?

mrxz:
> Probably good to go, but can I test on Quest 3 and AVP tonight to make sure?

FYI, I've tested it on a Quest 3 myself, but more testing never hurts. :-)

asundqui:
Looks like your code cleans things up and makes it simpler :). I checked on AVP additionally and everything checks out. Let's merge this yes @dmarcos thanks guys!

dmarcos:
Thanks

Issue #66, opened 2025-06-24
Significant loss in .ksplat -> .spz  conversion


shrimbly:
Hi there, thanks for your work on this project! 

Just flagging that I have noticed some significant loss in converting .ksplat files to .spz using the web UI on sparkjs.dev and when using the CLI with default arguments. I'm seeing large gaps in the scene, and there are issues with the spatial bounds being limited even with default settings.

.spz
![Image](https://github.com/user-attachments/assets/0278c3c0-338e-4c23-b10a-c5471573b823)

original .ksplat:
![Image](https://github.com/user-attachments/assets/25f75ffa-bb93-4d68-bff4-c8737d21c590)

Here is the ksplat file I'm having issues with. 
https://drive.google.com/file/d/19KT-GpM8xW3hXY4iRVs7B5xd7srtu2Ph/view?usp=sharing 

If there is any more information you need from me to help diagnose this, or if there is something I can be doing to improve the results here, please let me know!

dmarcos:
probably a bug in .ksplat reader implementation. looking into it

asundqui:
@shrimbly we just merged a fix for the issue, thank you for finding and reporting it!

Could you please try updating to the latest main and see if it works for you now? Btw you can also drag-and-drop and load the .ksplat file directly in Spark's viewer, without converting to SPZ. I found that converting to SPZ reduces your file size to approximately 1/3 though so I can see why you'd want to convert it.

dmarcos:
@shrimbly for your convenience to test with the latest build. https://stackblitz.com/edit/spark-flipbook-animation-m3afgyqe

dmarcos:
@shrimbly is the issue fixed for you?

dmarcos:
This should be fixed in main. It will ship in 0.1.5 version next week. Closing for now. Can reopen if necessary

shrimbly:
> [@shrimbly](https://github.com/shrimbly) is the issue fixed for you?

Hi @dmarcos - yes the primary issue is resolved in your latest build! No more large sections missing from the scene so thanks for your efforts there. 

One other similar issue though, it looks like the SH is being lost during conversion. Here is the same scene in .ksplat with SH2:
![Image](https://github.com/user-attachments/assets/d6ef2243-18aa-48c5-87d8-53c754c24a54)

here is the converted .spz:
![Image](https://github.com/user-attachments/assets/f7ec18f0-e7d5-4a40-bf19-65f862e08387)

[...]

PR #83, opened 2025-06-30
Fix Spherical Harmonics not included in SPZ transcoding.

asundqui:
When transcoding from KSPLAT to SPZ it turns out the Spherical Harmonics were not being included correctly! From #66, the .ksplat loading was corrected but transcoding still failed.

This PR fixes that by decoding and passing it along as necessary.


dmarcos:
Thanks!

PR #90, opened 2025-07-03
Use OffscreenCanvas and WebGL2 context to decode webp images for SOGS

mrxz:
For SOGS support (#73) a dependency was introduced to decode the webp images to access the raw RGBA pixel data. This PR replaces the dependency with using a WebGL context to read back the pixel data. This has the benefit of utilizing the browser's built-in image decoding.

There is the possible downside that browser implementation differences could lead to different result or that WebGL limitations on the device (e.g. max texture size) could limit which files can be decoded.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "0.1.4",
       "license": "MIT",
       "dependencies": {
-        "@jsquash/webp": "^1.5.0",
         "fflate": "^0.8.2"
       },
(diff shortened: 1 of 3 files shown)


dmarcos:
Thanks. It looks good to me

dmarcos:
Always happy to see a dependency gone. Merging this. It looks just a clean replacement of the decoding logic

PR #91, opened 2025-07-03
Exclude local assets directory from linting

mrxz:
After adding a splat using the SOGS file format to the local asset directory for testing, the `biome` linter started reporting on formatting issues for the `meta.json` file. Since the local assets directory is excluded by the `.gitignore` file already, it only makes sense to exclude it from linting as well.


dmarcos:
thanks

Issue #93, opened 2025-07-03
Errors when reusing ArrayBuffer from Cache or by concurrent loading (`envmap` example regression)


mrxz:
The `envmap` example no longer loads, which can be reproduced with the hosted example: https://sparkjs.dev/examples/envmap/index.html The following error is logged:
```
Uncaught (in promise) DOMException: Worker.postMessage: attempting to access detached ArrayBuffer
```

A git bisect indicates that the regression was introduced with #73. After some debugging this is caused by the fact that the examples loads the same `fireplace.spz` asset twice (concurrently) resulting in the same `ArrayBuffer` being used. This buffer is transferred to the worker, which fails the second time.

The internal usage of `THREE.FileLoader` by the `SplatLoader` actually introduces two ways in which an `ArrayBuffer` might get reused: concurrent requests or through `THREE.Cache`. Since the `SplatLoader` effectively "consumes" the buffer, it should never process the same buffer twice.

asundqui:
Okay with #94  and #112  I think we have fixed this issue and all the regressions, thanks for flagging it! @dmarcos once we've deployed this and tested I guess we can probably close out this issue?

mrxz:
Closing as #94 and #112 are both part of version 0.1.6 and resolve the issue.

PR #94, opened 2025-07-03
Use `fetch` instead of `FileLoader` in `SplatLoader` to avoid reusing same arraybuffer

mrxz:
A somewhat crude workaround for #93. Instead of using the `FileLoader` it uses the Fetch API directly. Goal is to fix the regression of the `envmap` example. There are still some points that can be improved:

- The `onProgress` callback is no longer called
- The error could still occur with SOGS as the "extra files" are still loaded using the `fileLoader`
- Loading the same file/URL will always result in separate fetches (no caching or combining concurrent requests)

### DIFF
--- a/src/SplatLoader.ts
+++ b/src/SplatLoader.ts
@@ -6,7 +6,7 @@ import { withWorker } from "./splatWorker";
 import { decompressPartialGzip, getTextureSize } from "./utils";
 
 // SplatLoader implements the THREE.Loader interface and supports loading a variety
-// of differeng Gsplat file formats. Formats .PLY and .SPZ can be auto-detected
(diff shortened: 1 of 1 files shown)


asundqui:
Wow thanks for finding this issue and fixing it so quickly. I just PR and merged another addition to parse SOGS from within a .Zip archive, hopefully it doesn't collide with this. But I'd say let's use your fix! Agree we could improve things in the future... probably at some point the whole file loading etc needs a cleanup and revamp, lots of paths now as a result of different file format support.

asundqui:
@mrxz would it be possible to make this handle onProgress callback? What would that involve, having it collect an array of ArrayBuffers for each packet, and then merging them at the end? I guess I can live with not handling all errors perfectly right now...

PR #96, opened 2025-07-04
Fix Examples Page

alex-mcneilly:
I've add hash navigation to the examples page, so you can now link directly to an example and the iframe changes accordingly. The examples list is also scrollable on mobile now. Additionally, I've added a fullscreen button in the bottom righthand corner next to the source code button. Now, developers can open the full example in a new tab without it being in an `<iframe>`.

Screenshot 2025-07-03 at 5 31 47 PM

### DIFF
--- a/examples.html
+++ b/examples.html
@@ -94,17 +94,28 @@
           border: none;
         }
 
-        .code-button {
+        .action-buttons {
           position: fixed;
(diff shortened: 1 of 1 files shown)


dmarcos:
You rock 🤘 Thanks so much.

PR #100, opened 2025-07-08
PC SOGS decoding from .Zip archive

asundqui:
Implement PC SOGS decoding from .zip file archive.
Updates viewer and editor to accept .zip files.

Reuses the existing dependency `fflate` decode a .zip file containing `meta.json` and WEBP image files referenced from it. Note that this works whether `meta.json` and the images are at the root of the archive, or if they are contained within some `subfolder/meta.json` - the decoder will grab the prefix of the first `meta.json` it sees and uses image files relative from that.

Here's a scan I did you can try: 
[canapes-mcmc-250k-30k.zip](https://github.com/user-attachments/files/21111735/canapes-mcmc-250k-30k.zip)

Any folder output zipped from the PlayCanvas SOGS package should work, and is a nicer interface to work with because you can drag-and-drop SOGS files into the viewer/editor.
image

### DIFF
--- a/examples/editor/index.html
+++ b/examples/editor/index.html
@@ -63,7 +63,7 @@
   <canvas id="canvas" tabindex="0"></canvas>
   <div id="main-gui"></div>
   <div id="second-gui"></div>
-  <input id="file-input" type="file" accept=".ply,.spz,.splat,.ksplat" multiple="true" style="display: none;" />
+  <input id="file-input" type="file" accept=".ply,.spz,.splat,.ksplat,.zip" multiple="true" style="display: none;" />
   <script type="module">
     import * as THREE from "three";
     import { OrbitControls } from "three/addons/controls/OrbitControls.js";
@@ -154,7 +154,7 @@
             loadFiles(urls);
             guiOptions.loadFromText = ""; // Clear after loading
           } else {
-            alert("No valid URLs found in text. URLs must start with http:// or https:// and end with .ply, .spz, .splat, .ksplat, or .json");
+            alert("No valid URLs found in text. URLs must start with http:// or https:// and end with .ply, .spz, .splat, .ksplat, .json, or .zip");
           }
         }
       },
@@ -683,7 +683,8 @@
         file.name.toLowerCase().endsWith('.ply') ||
         file.name.toLowerCase().endsWith('.spz') ||
         file.name.toLowerCase().endsWith('.splat') ||
-        file.name.toLowerCase().endsWith('.ksplat')
+        file.name.toLowerCase().endsWith('.ksplat') ||
+        file.name.toLowerCase().endsWith('.zip')
       );
 
       if (splatFiles.length > 0) {
@@ -693,7 +694,7 @@
 
     // Parse URLs from text (handles single URLs, multiple lines, mixed content)
     function parseURLsFromText(text) {
-      const supportedExtensions = [".ply", ".spz", ".splat", ".ksplat", ".json"];
+      const supportedExtensions = [".ply", ".spz", ".splat", ".ksplat", ".zip", ".json"];
       const urls = [];
 
       // Split by lines, commas, and semicolons
--- a/examples/viewer/index.html
+++ b/examples/viewer/index.html
@@ -347,7 +347,7 @@
       </div>
       <div class="drop-zone">Drag and drop a splat file here</div>
       <label for="file-input" class="button upload"><img class="upload-icon" src="upload-icon.svg" alt="upload" />Choose file</label>
-      <input id="file-input" class="hidden" accept=".ply,.spz,.splat,.ksplat" type="file" />
+      <input id="file-input" class="hidden" accept=".ply,.spz,.splat,.ksplat,.zip" type="file" />
       <form class="url-form">
         <input class="url-input" type="text" placeholder="Copy splat URL" />
       </form>
--- a/src/SplatLoader.ts
+++ b/src/SplatLoader.ts
@@ -1,3 +1,4 @@
+import { unzipSync } from "fflate";
 import { FileLoader, Loader, type LoadingManager } from "three";
 import { PackedSplats } from "./PackedSplats";
 import { SplatMesh } from "./SplatMesh";
@@ -138,6 +139,7 @@ export enum SplatFileType {
   SPLAT = "splat",
   KSPLAT = "ksplat",
   PCSOGS = "pcsogs",
+  PCSOGSZIP = "pcsogszip",
 }
 
 export function getSplatFileType(
@@ -155,6 +157,14 @@ export function getSplatFileType(
       return SplatFileType.SPZ;
     }
     // Unknown Gzipped file type
+    return undefined;
+  }
+  if (view.getUint32(0, true) === 0x04034b50) {
+    // PKZip file
+    if (tryPcSogsZip(fileBytes)) {
+      return SplatFileType.PCSOGSZIP;
+    }
+    // Unknown PKZip file type
     return undefined;
   }
   // Unknown file type
@@ -218,7 +228,7 @@ export type PcSogsJson = {
     maxs: number[];
     files: string[];
   };
-  shN: {
+  shN?: {
     shape: number[];
     dtype: string;
     mins: number;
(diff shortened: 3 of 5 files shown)


61cygni:
Awesome. Thank ou!

vv4dvv:
Hi, does the method of loading osgs only support  .zip  file, and not support loading models  of multiple webp image file at the same time?

After comparative tests, when using https://playcanvas.com/viewer to load models with the same webp files, it takes many times longer for us to load sutro.zip. Can we also support both zip and multi-file loading (like from meta.json?)

Reply from AI：The core reasons for the slow loading speed of zip archives are: the loss of the advantage of parallel downloading of multiple files, coupled with the additional CPU overhead and waiting time caused by decompression. For models composed of multiple webp images, a more efficient approach is usually to load independent webp files directly (taking advantage of parallel downloading and native decoding) rather than packaging them into a zip for loading via JS decompression.

dmarcos:
@vv4dvv Both .zip files and meta.json are supported by Spark. Give it a try and let us know how it goes.

vv4dvv:
@dmarcos  
     sogs/index.html    Load in this document

```
// const splatURL = await getAssetFileURL("sutro.zip");   //ok!
    // const splatURL = './model/sutro.zip';   //ok!
    // const splatURL = './model/sutro'    //Unknown splat file type: undefined
    const splatURL = './model/sutro/meta.json'   //Failed to construct 'URL': Invalid base URL
 
```
Additionally, the file unzipped from sutro.zip cannot be loaded when dragged to https://sparkjs.dev/viewer/, 
but it can be quickly loaded on https://playcanvas.com/model-viewer

dmarcos:
@vv4dvv a working example of loading a `meta.json` file https://stackblitz.com/edit/sogs-tnvdhx6h?file=index.html%3AL49

reminder you have to serve files via web server. browsers cannot access your local files

vv4dvv:
@dmarcos   Maybe we're not talking about the same thing. The following screenshots may provide a clearer comparison: the church Gaussian model is twice as large, but it loads faster. For the reason, please refer to the first AI response.
OSGS-loader

dmarcos:
> @dmarcos Maybe we're not talking about the same thing. The following screenshots may provide a clearer comparison: the church Gaussian model is twice as large, but it loads faster. For the reason, please refer to the first AI response. OSGS-loader

Sorry I’m not sure I understand your request / question. Spark supports both sogs via a .zip file or a meta.json that loads the individual .webp images as illustrated in https://github.com/sparkjsdev/spark/pull/100#issuecomment-3108376721

You can use whatever option best fits your needs

vv4dvv:
@dmarcos  Thank you for your quick reply. I used the code you provided just now, but the speed is still slow because you're still reading a single zip file, which results in slow transmission. The ideal scenario is to be able to read an entire folder as well, so that files can be downloaded in parallel. In this case, the download time would be determined by the largest file, and the speed could be many times faster than that of the original zip file.
osgs2

dmarcos:
@vv4dvv See line 49. The example loads a `meta.json` file

https://stackblitz.com/edit/sogs-tnvdhx6h?file=index.html%3AL49

```
const sutro = new SplatMesh({ url: "https://sparkjs.dev/assets/splats/sutro/meta.json" });
```

vv4dvv:
`
 const sutro = new SplatMesh({ url: "https://sparkjs.dev/assets/splats/sutro/meta.json" `});`

@dmarcos
It's true that six webp files can be seen loading this way, but the loading isn't parallel, and the loading time remains unchanged.

It may be necessary to use Promise.all() for concurrent loading.
OSGS-loader3

mrxz:
@vv4dvv This seems to be a bug in the Chrome Dev Tools rendering the concurrent requests on top of each other. Not sure what is causing it, but by selecting each image request and looking at the "Timing" tab shows that they are queued and started in parallel.

Using the Performance tab confirms this as well:
image

So the images _are_ being loaded concurrently AFAICT. It's of course still possible that there's a difference in loading performance between Spark and PlayCanvas. To properly compare the two you'd have to load the same file(s) from the same host, otherwise you might end up benchmarking your connection to a specific server.

mrxz:
Did a bit more digging, what Chrome shows in the Overview part of the Network tab is based on a `connectionId`. Since the CDN used by sparkjs.dev uses HTTP/2.0 the concurrent requests go over a single multiplexed connection. Since they share the same `connectionId`, they'll be rendered on the same band. They do show separately in the Performance tab as shown above and in the Network tab when showing the Waterfall.

The CDN used in the PlayCanvas example uses HTTP/1.1 so each request is a distinct connection, shown as distinct bands in the overview.

Relevant code: https://chromium.googlesource.com/devtools/devtools-frontend/+/refs/heads/main/front_end/panels/network/NetworkOverview.ts#81

vv4dvv:
`This seems to be a bug in the Chrome Dev Tools rendering the concurrent requests on top of each other. Not sure what is causing it, but by selecting each image request and looking at the "Timing" tab shows that they are queued and started in parallel.`
@mrxz Let’s set aside factors like browser DevTools versions and comparisons with PlayCanvas loading for now. Just based on our project’s .webp loading behavior  . If the loading were truly parallel, the loading time should roughly correlate with file sizes — larger files taking slightly longer, smaller ones faster. However, this is not what we’re seeing
Then looking at the loading function itself, GPT’s analysis suggests that the loading process doesn’t appear to be parallel either

OSGS4

mrxz:
> Let’s set aside factors like browser DevTools versions and comparisons with PlayCanvas loading for now. Just based on our project’s .webp loading behavior . If the loading were truly parallel, the loading time should roughly correlate with file sizes — larger files taking slightly longer, smaller ones faster.

@vv4dvv It's hard to set it aside when the DevTools _do_ tell us that the images are loaded concurrently. The smaller files might take longer than larger files due to a multitude of reasons. If you want to compare to PlayCanvas, do a comparison loading the _same files_ from the _same server_ over the _same network_.

As for the "GPT analysis", it doesn't suggest anything. It only states that `fetchWithProgress` can be used in a way that does allow parallel loading and a way that would cause serial loading. If you check the source code, you'll find that the pattern used by spark is the parallel one.

dmarcos:
@vv4dvv spark is fetching the files in parallel from the network. See Promise.all in https://github.com/sparkjsdev/spark/blob/339a610ea52360bf12b72e0450605a3b212272ab/src/SplatLoader.ts#L115

Not sure AI analysis is correct

vv4dvv:
`If you want to compare to PlayCanvas, do a comparison loading the same files from the same server over the same network.`

Since PlayCanvas is not based on the Three.js engine, I'm not sure how to implement a fair same-domain comparison. My experience with PlayCanvas is mainly limited to using their editor.

As for the enthusiasm around OSGS, I believe it's largely due to its **multi-file loading speed**, which is said to outperform the SPZ format. However, in practice, it still doesn't seem to meet expectations.

Another limitation is that `meta.json` currently only works over the network — it would be great if **local file paths** could also be supported. shanks

dmarcos:
@vv4dvv it’s easy to compare in equal conditions. Host the example I shared on github pages: https://stackblitz.com/edit/sogs-tnvdhx6h?file=index.html%3AL49

Load with playcanvas the same meta.json from the example also host on github pages.

You can load local files but have to serve them via a web server on your machine. Browsers can’t access local files directly for security / privacy reasons. It’s not a Spark / playcanvas or any other front end code limitation.

vv4dvv:
` it’s easy to compare in equal conditions`
@dmarcos  Could you provide a comparison demo that loads the `sutro` model using **PlayCanvas**, hosted on the same server as your existing case, so we can compare performance?

Also, in the case where we loaded a local relative path from the server, which was resolved with the help of AI, we can now clearly see **parallel loading lines (blue)** in the browser network panel. The load times now roughly correspond to the file sizes, which is exactly what we expected.

@mrxz  Does this mean that **the previous issue**—where loading didn't appear to be parallel and timing didn't match file size—might have been caused by **your server setup**?


OSGS6

dmarcos:
> @dmarcos Could you provide a comparison demo that loads the `sutro` model using **PlayCanvas**, hosted on the same server as your existing case, so we can compare performance?

Sorry. I’m not familiar with Playcanvas. It shouldn’t be difficult to put together with the pointers I gave you. I also recommend doing the exercise yourself to better understand the moving parts

- Host on github pages
- Use https://stackblitz.com/edit/sogs-tnvdhx6h?file=index.html%3AL49 code for the spark version
- Load https://sparkjs.dev/assets/splats/sutro/meta.json in both.

We already know both Spark and Playcanvas load files in parallel so there should not be any difference in network fetching times.

vv4dvv:
`We already know both Spark and Playcanvas load files in parallel so there should not be any difference in network fetching times.`
@dmarcos  I found that after uploading to the server, it again showed as a single unified band. Just as @mrxz mentioned, both the local server and PlayCanvas use HTTP/1.1, while most cloud servers use HTTP/2. Thank you all — this issue is finally clarified.

vv4dvv:
I think I have found the answer. Although we all agree that the code downloads in parallel, the observed phenomenon is that at any given time, only one `.webp` file's size is increasing, which feels like a limitation. The HTTP/2 protocol and the fetch method are more modern, but why does the PlayCanvas example not use them? Instead, it falls back to HTTP/1.1 and XHR methods, while other resources still use HTTP/2.

Below is the AI's explanation:
In HTTP/2, browsers implement priority scheduling for large files and concurrent image requests.
Although HTTP/2 supports multiplexing—allowing multiple requests to be sent over a single connection concurrently—
the browser does not always actively transfer multiple large files at the same time.
Instead, it determines which requests to prioritize based on factors such as:
content type，decoding cost，memory pressure，stream priority
As a result, you may observe that all requests are sent, but only one resource is actively receiving data at a time.
🌐 This behavior is known as request stream shaping, priority throttling, or multiplexed stream prioritization.

dmarcos:
@vv4dvv  What protocol (http1 or 2) the browser uses depends on the server where assets are hosted and not the library or method used. You can host your assets in whatever server yields the best results for your use case.

I recommend getting familiarized with the nuances of multiplexing / parallelism (AI not ideal to truly understand). Ultimately, download time is limited by bandwidth that can be divided / distributed in different ways across requests

At the end, you want to measure page load times (time until user can interact with content) that it’s what the user perceives.

Issue #77, opened 2025-06-29
The visible prop only has effect if used in <SparkRenderer> and not on <SplatMesh>


joaomelorodrigues:
<img width="418" alt="Image" src="https://github.com/user-attachments/assets/8cf33ef3-14e4-4f89-9c81-4782ab5afd8c" />

dmarcos:
cc @dylanhu7

dylanhu7:
Thanks for creating this issue. The `visible` prop definitely should work for toggling visibility when given to `SplatMesh`. We will be looking into this.

asundqui:
The SplatMesh should handle the `.visible` property, but it's possible there is an issue with it not toggling the `needsUpdate` flag correctly.. @cyango do you have an example of this happening, video or anything?

joaomelorodrigues:
> The SplatMesh should handle the `.visible` property, but it's possible there is an issue with it not toggling the `needsUpdate` flag correctly.. [@cyango](https://github.com/cyango) do you have an example of this happening, video or anything?

Here:

https://github.com/user-attachments/assets/f3f6583e-36d1-44ad-8f55-3bdc35db0b4b

dmarcos:
Fixed by https://github.com/sparkjsdev/spark/pull/101

PR #101, opened 2025-07-08
Fix visible flag bug in SplatGenerator/SplatMesh

asundqui:
This PR should hopefully fix the issue #77 because in fact the .visible flag was not being respected by SparkRenderer! In SparkRenderer.compileScene we gather all SplatGenerators because we want to be able to run all their updates functions, irrespective of their .visible flag. This PR updates compileScene to gather both all generators and only the visible subset (including parent sub-tree visibility effects). Fortunately setting `.opacity=0` has been available as a workaround, but this was quite an unfortunate oversight!


dylanhu7 (approved):
Nice!

dmarcos:
Thanks!

PR #102, opened 2025-07-08
DRAFT: Adjust focal and ndcOffset computations

mrxz:
This PR is mostly for demonstration purposes. The changes bring the behaviour in line with SuperSplat, generally resulting in sharper results. However, I believe these changes are technically not correct, despite the subjectively better visuals.

It's worth noting the Mkkellogg's renderer has a [`focalAdjustment` parameter](https://github.com/mkkellogg/GaussianSplats3D/blob/2dfc83e497bd76e558fe970c54464b17b5f5c689/src/splatmesh/SplatMaterial3D.js#L206-L207) that also tweaks this behaviour, quoting the docs:

>  `focalAdjustment`: Hacky, non-scientific parameter for tweaking focal length related calculations. For scenes with very small gaussians & small details, increasing this value can help improve visual quality. Default value is 1.0.

So it might be worthwhile to consider a similar biased/subjective option. Of course, some additional scrutiny to the logic to ensure the current implementation _is_ correct would be appreciated. Given these changes match the SuperSplat behaviour, one or the other must be doing something "wrong" (deliberate or not).

### DIFF
--- a/src/shaders/splatVertex.glsl
+++ b/src/shaders/splatVertex.glsl
@@ -113,7 +113,7 @@ void main() {
 
     // Compute the Jacobian of the splat's projection at its center
     vec2 scaledRenderSize = renderSize * renderScale;
-    vec2 focal = 0.5 * scaledRenderSize * vec2(projectionMatrix[0][0], projectionMatrix[1][1]);
+    vec2 focal = scaledRenderSize * vec2(projectionMatrix[0][0], projectionMatrix[1][1]);
     float invZ = 1.0 / viewCenter.z;
     vec2 J1 = focal * invZ;
     vec2 J2 = -(J1 * viewCenter.xy) * invZ;
@@ -177,7 +177,7 @@ void main() {
 
     // Compute the NDC coordinates for the ellipsoid's diagonal axes.
     vec2 pixelOffset = eigenVec1 * scale1 + eigenVec2 * scale2;
-    vec2 ndcOffset = (2.0 / scaledRenderSize) * pixelOffset;
+    vec2 ndcOffset = pixelOffset / scaledRenderSize;
     vec3 ndc = vec3(ndcCenter.xy + ndcOffset, ndcCenter.z);
 
     vRgba = rgba;


asundqui:
This is a very interesting finding. TBH I don't know for certain what the "correct" value is anymore... It seems like it should perhaps be 0.5 and 2x because NDC space is -1..1... @mrxz do you have an opinion here?

@bmild do you have any knowledge/memory of what the correct behavior should be?

Would it make sense to have a similar `focalAdjustment` to allow people to change this setting globally? We still have the question of what the "default" value should be then, but I think it's a good idea to have knobs people can turn to tune Spark to their liking.

dmarcos:
Ground truth should be the images used to train the splat? For some of the splats in the examples we have the images or could ask for them if useful to have. Having a test set to compare with other solutions and detect changes regressions would be useful

mrxz:
> It seems like it should perhaps be 0.5 and 2x because NDC space is -1..1... @mrxz do you have an opinion here?

Assuming the `pixelOffset` is indeed in pixels, the current logic of `(2.0 / scaledRenderSize) * pixelOffset` seems correct to me. The center of the screen would be (0,0) in NDC, so anything half a screen over should become 1 not 0.5.

Similarly the projection matrix has (assuming a symmetric view volume) `near / right` and `near / top` in `projectionMatrix[0][0]` and `projectionMatrix[1][1]` respectively. Factoring in the viewport dimensions in pixels we'd use half the width and height, as is currently done by the 0.5 factor.

So I don't really see why these changes would be correct, yet the results are simply sharper (though not particularly aliased), align with SuperSplat and subjectively look nicer. Really feels like I'm missing something.

asundqui:
I agree with your calculation... seems like it should be correct. I think the best "authority" may be using the training values... I think the "gsplat" library could possibly be considered the most defacto standard?

In their repo I see this: https://github.com/nerfstudio-project/gsplat/blob/73fad53c31ec4d6b088470715a63f432990493de/gsplat/cuda/csrc/ProjectionEWASimple.cu#L44
```
float fx = Ks[0], cx = Ks[2], fy = Ks[4], cy = Ks[5];
```
Connected to that in `persp_proj` I see the following: https://github.com/nerfstudio-project/gsplat/blob/73fad53c31ec4d6b088470715a63f432990493de/gsplat/cuda/include/Utils.cuh#L528
```
    mat3x2 J = mat3x2(
        fx * rz,
        0.f, // 1st column
        0.f,
        fy * rz, // 2nd column
        -fx * tx * rz2,
        -fy * ty * rz2 // 3rd column
    );
```
So even though the gsplat library doesn't have the 2.0 and 0.5 factors in there, I'm pretty sure the camera intrinsics in `K` are `width`,`height`, and `-0.5*width`, `-0.5*height` or similar, in other words converting from view space to pixel space. So it seems we should also be doing our calculation in pixel space, which involves 2.0 and 0.5 because we're going through NDC.

I feel inclined to believe that Spark + MKellogg to it "more correctly" here... although maybe it would still be nice to provide a `focalAdjustment` or similar parameter for users who want that degree of control?

mrxz:
Reaching the same conclusion when looking at [graphdeco-inria/diff-gaussian-rasterization](https://github.com/graphdeco-inria/diff-gaussian-rasterization).

Width and height are halved for the focal point ([rasterizer_impl.cu#L222-L223](https://github.com/graphdeco-inria/diff-gaussian-rasterization/blob/59f5f77e3ddbac3ed9db93ec2cfe99ed6c5d121d/cuda_rasterizer/rasterizer_impl.cu#L222-L223)):
```cuda
const float focal_y = height / (2.0f * tan_fovy);
const float focal_x = width / (2.0f * tan_fovx);
```

And the resulting NDC coordinate is converted into pixel space, hence no 2.0 factor here ([forward.cu#L232-L233](https://github.com/graphdeco-inria/diff-gaussian-rasterization/blob/59f5f77e3ddbac3ed9db93ec2cfe99ed6c5d121d/cuda_rasterizer/forward.cu#L232-L235)). The computed radius (in pixels) is simply used to determine the bounding rectangle, whereas we need to convert the pixel offset into NDC.
```cuda
float my_radius = ceil(3.f * sqrt(max(lambda1, lambda2)));
float2 point_image = { ndc2Pix(p_proj.x, W), ndc2Pix(p_proj.y, H) };
uint2 rect_min, rect_max;
getRect(point_image, my_radius, rect_min, rect_max, grid);
```

> I feel inclined to believe that Spark + MKellogg to it "more correctly" here... although maybe it would still be nice to provide a focalAdjustment or similar parameter for users who want that degree of control?

It would definitely be nice to offer users some control over it. Difficulty is that since it's subjective, it's hard to say what the best approach is. From my limited testing simply reducing `preBlurAmount` doesn't quite cut it and it quickly introduces aliasing.

Instead of replicating the behaviour of `focalAdjustment` there's also the option to use (or rather abuse) the recently added `renderScale` option for this. Again subjectively, but the `focalAdjustment` seemed to have a tendency to "erode" the Gaussians way more when getting close compared to increasing the `renderScale`. On top of that it will be easy to help people comparing or moving from SuperSplat since a `renderScale` of `2.0` can be used.

Obviously it would warrant renaming the property and documenting it. We could still call it `focalAdjustment` for example. Although it doesn't quite do the same as the mkkellogg equivalent, the purpose is the same.

dmarcos:
Can we close this in favor of https://github.com/sparkjsdev/spark/pull/113 or there's anything else to do?

dmarcos:
It looks https://github.com/sparkjsdev/spark/pull/113 replaces this. Closing for now. Can reopen if necessary

Issue #87, opened 2025-07-02
docs: missing navigation on small/medium viewports


joestrouth1:
The docs feel frustrating to navigate in anything other than a large browser window. When I view the Spark docs there, I can see a list of docs pages on the left and a table of contents for the current page on the right.

![Image](https://github.com/user-attachments/assets/9a12460b-e2fd-45c7-9aa0-703f8bc91557)

On smaller viewports like a half-width window, tablet, or phone, those disappear. The docs nav on the left hides at <1220px wide and the table of contents goes away at <960px wide. The logo also disappears.

![Hidden nav at medium viewports](https://github.com/user-attachments/assets/53bc07bf-ff44-468e-b86b-dbec016d0523)

![Hidden table of contents at small viewports](https://github.com/user-attachments/assets/39d71fb6-c01f-4963-b7a9-7cdf1b524cc6)

I can see there is a drawer navigation by poking around in dev tools, but there's no obvious "Menu" button to make it appear. If I check a hidden input, it slides in:

![Image](https://github.com/user-attachments/assets/348e40e3-4b67-475d-9c4f-2e435c6336d2)

I read through the [Material for MKDocs docs section on Navigation](https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/) and didn't see any options related to this responsive behavior. There are options to hide the nav/ToC but I only saw those used on the homepage. Removing them from `index.md` did not help. Didn't see anything fishy in the extra `spark.css` stylesheet or `mkdocs.yml` either.

--- 

I don't see anything that looks like a template file for docs pages or the site header. If MKDocs or the material theme the site uses supports that, one could add a menu/ToC button there. I know MKDocs supports loading extra javascript, where one could create a `button` that appears only on small viewports and toggles the `input[type=checkbox]` responsible for controlling the drawer.

The drawer as it exists today doesn't use the logo/wordmark visible in the global nav, but would at least offer navigation.

### Expected

I can navigate within the article I'm viewing and between docs pages on small screens or when splitting my display between docs+project.

### Actual

I must click 'Docs' in the header to get back to the Getting Started page, then scroll to the bottom and find the link I'd like to go to.

### Steps to reproduce

1. Navigate to the [SparkRenderer docs page](http://localhost:8000/docs/spark-renderer/) in a viewport <1220px wide
2. Try to navigate to the [SparkViewpoint docs page](http://localhost:8000/docs/spark-viewpoint/)

[...]

PR #105, opened 2025-07-09
docs: Add menu drawer toggle to header on smaller viewports

joestrouth1:
See #87 

Adds a 'Menu' button the the site header which opens the navigation drawer when clicked. 

\>=1220px wide (unchanged):

image

Tablet:

image

It needs work on phones. The header is already too wide for small (iPhone SE) screens and adding a menu/drawer toggle worsens the problem. 

image

Some options for shrinking the width of header contents:
- Swap 'Menu' for an icon
- hide 'Docs' link since the menu drawer lists all the docs pages
- hide GH/discord icons in sm-md header
  - maybe override the drawer or footer to show them there


dmarcos:
Oh wow. this is awesome. It looks good. Gonna give it a try tomorrow before merge. Thanks so much for the effort

dmarcos:
iphone SE is as small as phones can be these days and you have space from screenshot above. reduce left margin / padding of the magnifiying lense and use a "menu icon" instead of menu word

joestrouth1:
Most recent changes:
- swaps 'Menu' for a material icon
- reduces inline padding on the `header.md-header` by half from 20px to 10px on small viewports
- reduces padding around the Search and Menu buttons by half from 8px to 4px on small viewports
    - The size of each button decreases to 32px square which is between WCAG's minimum target size of 24px and their enhanced recommendation of 44px
- reduces the gap between items in the `.md-custom-header-right` section from `1rem` to `0.5rem` on small viewports
    - moves the GH icon into the `.md-custom-header-right` and removes `margin-right` (relying on `gap` instead)

### iPhone SE:
image
image

## iPhone 14 Pro Max:
image

### iPad Mini:
image

### iPad Pro / laptop class:
image

dmarcos:
Thanks so much. Icons should be probably spread more evenly and centered in the header in mobile screens. Also can we put the spark logo on the mobile menu vs the default one?

LeeBurton:
I think it looks a bit odd with the Search icon before the Menu icon. I think it would look better with the Search icon positioned after all menu items and before the Discord icon. This is then also in line with how other websites are structured. Plus, the most used items should be on the left and lesser used further right. Menu is more important than the Search.

PR #106, opened 2025-07-09
DRAFT. Prevent default wasm-pack behavior when fetching .wasm files via new URL that seems to confuse Webpack (fix #97, fix #95)

### LINKED ISSUE #95: Nextjs import / webpack
It (0.1.5) seems to fail with webpack 5 / next js 15.3.4.

I really have no idea why after diving into it a bit.  Tested it on a vanilla install.

3x
<w> [webpack.cache.PackFileCacheStrategy] Serializing big strings (114kiB) impacts deserialization performance (consider using Buffer instead and decode when needed)
Failed to compile.

./node_modules/@sparkjsdev/spark/dist/spark.module.js
Module not found: Invalid generator object. Asset Modules Plugin has been initialized using a generator object that does not match the API schema.
 - generator has an unknown property 'filename'. These properties are valid:     
   object { binary?, dataUrl? }
   -> Generator options for asset/inline modules.

### LINKED ISSUE #97: NextJS / Webpack Compatability Issues
## Problem
Spark does not work with the latest NextJS version (`15.3.4`).

  SparkJS uses Webpack-specific syntax for loading WASM files, specifically `new URL()` with `import.meta.url` which is incompatible with Next.js's webpack configuration. I get the following error:

```Module not found: Invalid generator object. Asset Modules Plugin has been initialized using a generator object that does not match the API schema.```

**My attempts at fixing it in my file (unsuccessful)**
1. Configuring the `next.config.ts` to handle the WASM imports
2. Creating a wrapper around SparkJS that disables server-side rendering
3. Creating a custom loader that intercepts WASM loading

## To Reproduce
1. Start a new repo with `npx create-next-app@latest`
2. Install SparkJS and try to render any `.ply` file with `SplatMesh`

## References for inspiration
1. `gsplat.js` does not use the `new URL(..., import.meta.url)` pattern that SparkJS uses. Instead, it loads WASM as a regular ES module dependency or fetches it via HTTP, which is compatible with most modern bundlers, including Next.js.
2. `@zappar/three-gaussian-splat` either inlines the WASM or fetches it as a static asset, not using the new URL(..., import.meta.url) pattern.

dmarcos:
Don't merge. tentative solution. needs verification.

### DIFF
--- a/src/SplatMesh.ts
+++ b/src/SplatMesh.ts
@@ -1,6 +1,7 @@
 import * as THREE from "three";
 
 import init_wasm, { raycast_splats } from "spark-internal-rs";
+import wasmURL from "spark-internal-rs/spark_internal_rs_bg.wasm?url";
 import { PackedSplats } from "./PackedSplats";
 import { type RgbaArray, readRgbaArray } from "./RgbaArray";
 import { SplatEdit, SplatEditSdf, SplatEdits } from "./SplatEdit";
@@ -252,7 +253,9 @@ export class SplatMesh extends SplatGenerator {
   static dynoTime = new DynoFloat({ value: 0 });
 
   static async staticInitialize() {
-    await init_wasm();
+    const response = await fetch(wasmURL);
+    const wasmBytes = new Uint8Array(await response.arrayBuffer());
+    await init_wasm({ wasm: wasmBytes });
     SplatMesh.isStaticInitialized = true;
   }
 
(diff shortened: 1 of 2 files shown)


dmarcos:
no longer needed. This is [the recommended workaround](https://github.com/sparkjsdev/spark-react-basic?tab=readme-ov-file#using-webpack)

PR #109, opened 2025-07-11
Add sh1, sh2 support to sogs

lucasoptml:
pcsogs loader assumed sh3, this supports sh1 and sh2 too.

tested on sh2 pinecone-sogs.


61cygni:
thank you!

PR #110, opened 2025-07-11
Expand internal splat scale range to e^[-12,9] from e^[-9,9]

asundqui:
This PR expands the minimal splat scale range from e^-9 (0.0001) to e^-12 (0.000006). This allows for 20x thinner structures to be represented. The "zero cutoff" threshold is lowered to e^-30 because the test file showed structures below e^-20, which I previously believed we'd never seen.

I tested lower-end ranges e^-9, e^-10, e^-11, e^-12, and e^-13 and found that below e^-12 there were no perceivable differences. There is of course the chance that a user could upload a file with an even smaller scale and even finer structures, but in order to encode each scale_xyz in a uint8 we need to make a compromise in allowed range of scale vs. resolution of each scale step. The range e^[-12,9] provides e^(21/254) =1.0862x steps between each quantized scale. Previously the range e^[-9,9] had 1.0734x steps, so the difference between the previous encoding is minimal.

Note that the SPZ file format encoding itself limits the range to e^[-10,6], so if the user has SPZ files they still won't be able to get structures below e^-10 in scale.

Before change:
sutro_-9+9

After change:
sutro_-12+9

This PR also fixes background color selection in editor, which had a bug leading to colors being off.


61cygni:
This is great, thank you.

PR #112, opened 2025-07-12
Implement SplatLoader onProgress callback, misc. improvements to examples

asundqui:
Update SplatLoader to handle progress updates, both for main file and any extra files, aggregating totals.

Misc improvements to examples:
- Added missing <html> tags
- Any examples that loaded the same splat file multiple times were modified to load it once as a PackedSplats, then using that to instantiate multiple SplatMeshes
- Fixed envMap example
- SOGS example: Set high-DPI pixel ratio. Decrease near clipping plane to 0.01 from 0.1
- Changed default focal distance for splat texture

I tested all examples and they all work correctly.

### DIFF
--- a/examples/envmap/index.html
+++ b/examples/envmap/index.html
@@ -1,4 +1,5 @@
 <!DOCTYPE html>
+<html>
 
 <head>
   <meta charset="utf-8">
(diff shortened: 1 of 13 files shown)


61cygni:
Amazing, thank you!

examples/envmap/index.html
```diff
@@ -55,13 +56,14 @@
     const spark = new SparkRenderer({ renderer });
 
     const splatURL = await getAssetFileURL("fireplace.spz");
-    const background = new SplatMesh({ url: splatURL });
+    const packedSplats = new PackedSplats({ url: splatURL });
+    const background = new SplatMesh({ packedSplats });
```
dmarcos, line 60: Cannot longer load same splat via URL multiple times with the old API?

asundqui: You can, it will work, but it will fetch the data from the URL twice, decode it twice, upload all the data twice, etc. For a little while it was broken before #94 fixed it, so it works again. But I'd say it's a non-idiomatic way to do things and wasteful, I figure should show how to do it more efficiently in the examples.

PR #113, opened 2025-07-12
Rename SparkRenderer.renderScale to focalAdjustment. 

### LINKED ISSUE #99: Same file looks blurrier and with some artifacts compared to Supersplat
Sparks looks great! but unfortunately, it degrades the quality compared to supersplat web viewer

Image

You can get the test file from https://wetransfer.com/downloads/493bfd6235042deb9aa6120c6bcd8c3020250707075148/848258e49f8610cab6ecb4457b15fdf120250707075308/923e87?t_exp=1752133908&t_lsid=9dbdce9f-aace-40a4-8933-efdb82931da1

Regards, Víctor

asundqui:
Also added to constructor options for initialization and to docs. Included note about how it's similar to the same `focalAdjustment` parameter in MKellogg, and that a value of 2.0 matches the calculations in the PlayCanvas engine.

This PR subsumes #102 by providing a parameter that people can use to adjust to their liking. This potentially fixes #99 but it will be up to the user to decide if these tweaks are enough to match their expectation.

### DIFF
--- a/docs/docs/spark-renderer.md
+++ b/docs/docs/spark-renderer.md
@@ -39,6 +39,7 @@ const spark = new SparkRenderer({
   apertureAngle?: number;
   falloff?: number;
   clipXY?: number;
+  focalAdjustment?: number;
   view?: SparkViewpointOptions;
 });
 ```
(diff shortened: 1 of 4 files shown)


dmarcos:
good stuff! thanks!

PR #123, opened 2025-07-15
Use JSDoc for documenting `SparkRenderer` and `SparkViewpoint` options

mrxz:
The source code documents many of the constructor options of the `SparkRenderer` and `SparkViewpoint`. However these are written using multiple single-line comments, effectively making them for internal use only. By using JSDoc the typescript compiler picks them up and retains them when emitting the type definitions. This allows users importing Spark in their projects to see this documentations in their IDE for both JS and TS projects.

This PR updates the documentation comments for the `SparkRendererOptions` and `SparkViewpointOptions`. Ultimately more parameters, methods, classes and functions can be documented this way. If needed it can also be used to reference external resource like the online documentation of Three.js or Spark itself.

Here's an example screenshot showing it in effect in VS code:
image

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -78,62 +78,101 @@ THREE.Scene.prototype.onBeforeRender = function (
 };
 
 export type SparkRendererOptions = {
-  // Pass in your THREE.WebGLRenderer instance so Spark can perform work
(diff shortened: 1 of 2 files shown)


dmarcos:
looks great to me

PR #127, opened 2025-07-21
Remove `anyhow::anyhow` dependency to reducing bundle size

mrxz:
Turns out that the `anyhow::anyhow` dependency bloats the resulting .wasm file quite a bit. It was only used once and can easily be replaced with a simple `Result<u32, String>`.

This results in the following file size savings:
| File | Before | After |
|--|--|--|
|`spark.module.min.js`| 653.09 kB | 492.99 kB |
|`spark.cjs.min.js`| 553.30 kB | 393.19 kB |


dmarcos:
Thanks. This needs rebase

mrxz:
@dmarcos Rebased

dmarcos:
thanks

Issue #104, opened 2025-07-08
Rasterization when using sh3


murcje:
Hi!

I am experiencing a rasterization issue when using spark on my platform. The splat looks good in https://sparkjs.dev/viewer/ but on my platform I get this horizontal rasterization.

![Image](https://github.com/user-attachments/assets/1c2ee660-106d-4eee-804e-7d482a41b30e)

When I use an sh2 file there is no raster.

![Image](https://github.com/user-attachments/assets/30069ba2-2603-4a45-b310-306187b0cc5c)

Here is the ply: https://storage.googleapis.com/voluma-assets/voluma/testestes/PM_Lykke_250KSplats_Cropped_Compressed.ply

And here is the ply on my site, with rasterization:

https://voluma.ai/embed/voluma/testestes/lykke

dmarcos:
Thanks for filing. We'll look into it

mrxz:
@murcje These "rasterization lines" aren't due to spherical harmonics but are caused by quantization. Looking at the provided .ply file the splats all have Y coordinates < -210, which is quite a way away from the origin. I needed to drag the camera a lot when loading it in the viewer, but saw the same artefacts. Could it be that when you tested it in the viewer you used a different file?

In any case, I would recommend adjusting the splat file so that the splats are near/around the origin and also to try and keep your `SplatMesh` close to the origin of your scene as well. That should yield the best results and avoid most quantization issues.

murcje:
@mrxz I believe you are right. The splat was created by a client of ours. It is positioned at about +210 on the y-axis.
 I placed the group of gaussians on the origin and the artefacts are gone. 

Thank you for taking time to check!!

Issue #81, opened 2025-06-30
Discrepancy of text brightness/sharpness Supersplat viewer and Sparkjs


meditation-error:
I have this scene using sparkjs: https://voluma.mediamarc.nl/embed/jack/x/watch
and this one is in supersplat viewer: https://rr3d.dk/Upload/3dgs/watch/Watch_MCMC_1M_150KSteps_Compressed/index.html

Supersplat does seem to do some custom color correction I assume, but I was wandering, how could the text be so bright ?
Especially the numbers around the edge, but also the text 'Alpha slowmaster Professional' is very bright. Almost as if it has a higher sh level with more/different data. I tried asjusting, blur, brightness, contrast in spark, but the text stays in much lower contrast with the backplate.

Mkkellogg's viewer is showing roughly the same as sparkjs, and maybe it's just some custom improvement by the supersplat viewer, but I wander if anyone has any insight how this brightness is achieved and could it maybe be recreated in spark ?

The file:  https://assets.voluma.ai/rr3d/scene/Watch_MCMC_1M_150KSteps_Compressed.ply

![Image](https://github.com/user-attachments/assets/7e79948a-053f-4521-bcf8-a23fd0d358b4)

dmarcos:
Thanks for providing details. We’ll look into it

dmarcos:
@meditation-error we are investigation. By any chance you have access to the images / video used to train the splat? Thanks

mrxz:
The difference in brightness seems to stem from the way the RGB values from the .ply file are handled. Both spark and mkkellogg decode the colour values and quantize the 0-1 range into 8-bit values. Supersplat on the other hand does the decoding in the shader and as a result can use values outside of the 0-1 range (brighter colours).

To illustrate, here's a comparison showing Supersplat without modification and with the same clamping applied. Note how this results in the same muted look:

![Image](https://github.com/user-attachments/assets/4afa45f8-c48b-49a1-b7b1-c8e9272cbca1)

meditation-error:
Thanks for looking into it @mrxz that sounds plausible.
Do you think Spark can support color decoding in the shader too at some point ?

@dmarcos Not sure if still needed, but here is the source material: https://drive.google.com/drive/folders/1btktki1IdDnHmiX0c-HzLn1FDpw8n7Cp?usp=sharing

[...]

Issue #36, opened 2025-06-06
Quality is low in comparison to other viewers


hybridherbst:
Thanks for starting Forge!

I made some quality comparisons and I can't get the Forge viewer to look nearly as good as the others, both for low and for high splat counts. 
The results are much blurrier (in both cases) and downright bad (for the high-splatcount version).

Is this known? Are you planning to improve quality?

# Screenshots
| Viewer | 10k splats | 1200k splats |
| - | - | - |
| Forge | <img height="300" alt="Image" src="https://github.com/user-attachments/assets/ba4240fb-41c3-4ebe-b79e-51068c25a392" /> | <img height="300" alt="Image" src="https://github.com/user-attachments/assets/9d9ce9d6-0f91-4e77-94cc-2e5e508e0000"/> |
| SuperSplat | <img height="300" alt="Image" src="https://github.com/user-attachments/assets/cbd85c2f-b3e6-4d49-8950-e95934c21e6a" /> | <img height="300" alt="Image" src="https://github.com/user-attachments/assets/cdebf3aa-91c9-4e6e-9f36-c752a9e06c8d" /> | 
| mkkellogg | <img height="300" alt="Image" src="https://github.com/user-attachments/assets/598fd267-3341-46b4-aa8d-424d3987a2b8" /> | <img height="300" alt="Image" src="https://github.com/user-attachments/assets/0788bc30-7780-49e2-b342-82eb63729251"/> | 

# Reproduction files

I attached the 10k splat file here: [AnisoLamp2.ply.zip](https://github.com/user-attachments/files/20627023/AnisoLamp2.ply.zip)

The 1200k splat file can be directly viewed:
- on Forge: https://forge.dev/viewer/?url=https%3A%2F%2Fd28zzqy0iyovbz.cloudfront.net%2F1b0b385d%2Fscene.compressed.ply
- on SuperSplat: https://superspl.at/view?id=1b0b385d

# Additional Info
In case it's related, I'm on a high-dpi screen – not sure if maybe DPI isn't correctly taken into account in the viewer?

dmarcos:
Thanks so much for taking the time to compare them all. Looking into it.

asundqui:
Hey there @hybridherbst , thank you for bringing this to our attention and doing all the work to show comparisons and get us the repro. We've looked into it and I think we have a fix!

Before fix:
[...]

PR #128, opened 2025-07-21
Draft: Expand splat represention to 32 bytes using two array texture targets

mrxz:
This PR is to explore the possibilities when extending the splat representation from 16 bytes to 32 bytes (2x16) using MRT. Vanilla three.js does not support MRT with 2d texture arrays out-of-the-box, but requires surprisingly few changes (see https://github.com/mrdoob/three.js/pull/30151 and https://github.com/mrxz/three.js/commit/e63a1af54b8a93b500814daae628d85aec4d127b). Hopefully we can get this implemented in Three.js upstream. In this PR I point to a three.js build with the needed changes, so anyone can try it out.

For simplicity the newly available space has only been used to increase the resolution of the center (X, Y, Z) and the colour values, resulting in the following layout:
| Channel | Contents |
|-------------|---------------|
|R1            | `center.x` (32f) |
|G1            | `center.y` (32f) |
|B1            | `center.z` (32f) |
|A1            | `scale.xyz` (3 x 8bit), unchanged from current representation, 8 bit unused |
|R2            | `rgba.r` (16 bit) + `rgba.g` (16 bit) |
|G2            | `rgba.b` (16 bit) + `rgba.a` (16 bit) |
|B2            | `encodeQuatOctXy88R8` (24bit), 8 bit unused |
|A2            | unused|

The main motivation behind was to address the following two prominent issues: 

- float16 quantization issues with splats away from origin (either in source file or moved in scenegraph): https://github.com/sparkjsdev/spark/issues/104, https://github.com/sparkjsdev/spark/issues/36#issuecomment-3074858674
- "dimmer" results due to clamping of colour channels: https://github.com/sparkjsdev/spark/issues/81, https://github.com/sparkjsdev/spark/issues/36#issuecomment-3074618885

Of course, there's plenty of unused bits, so the scales and rotation can be encoded with higher precision as well. For the colour channels I've opted for encoding them as 0-65535 converted to 0.0-257.0 in the shader. This is overkill, but can at least represent the values > 1.0, which with pre-multiplied alpha (also in this PR) brings back the brighter highlights. Ideally we pick a better representation that closely matches the possible input values, which might even be negative. (Though it's unclear to me if negative colour channel values contribute meaningfully to a splat model)

The biggest drawback is of course that memory consumption is doubled. And the impact on performance will have to be assessed properly. Since I only have a reliable profiling setup for the Quest 3 at the moment, I've measured the app time (CPU + GPU) for the webxr example in the repo. It increased slightly from 15.534ms (before) to 15.614ms (after).

Not surprising as the GPU metrics never indicated texture fetching to be the bottleneck on the Quest 3. So having to perform more texture fetches per vertex shader execution doesn't seem to be a problem. Though no idea if this generalizes to other TBDR GPUs.

To make the texture uploading easier, I've opted for representing the `packedArray` as two arrays, one for each texture. This is somewhat awkward representation (especially for the loaders). The alternative would be to tightly pack the splats in one array (as is currently the case) and split them into two when uploading to the GPU, which isn't ideal either...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "onchange": "7.1.0",
         "spark-internal-rs": "file:rust/spark-internal-rs/pkg",
         "stats.js": "^0.17.0",
-        "three": "^0.172.0",
+        "three": "git+https://github.com/mrxz/three.js.git#array-texture-mrt-build",
         "ts-node": "10.9.2",
         "typescript": "^5.7.3",
         "vite": "^6.0.11",
@@ -2623,9 +2623,8 @@
       }
     },
     "node_modules/three": {
-      "version": "0.172.0",
-      "resolved": "https://registry.npmjs.org/three/-/three-0.172.0.tgz",
-      "integrity": "sha512-6HMgMlzU97MsV7D/tY8Va38b83kz8YJX+BefKjspMNAv0Vx6dxMogHOrnRl/sbMIs3BPUKijPqDqJ/+UwJbIow==",
+      "version": "0.178.0",
+      "resolved": "git+ssh://git@github.com/mrxz/three.js.git#bc2d35e0b7c33c4fd9d3f7ef58a6765f9527adbf",
       "dev": true,
       "license": "MIT"
     },
--- a/package.json
+++ b/package.json
@@ -48,14 +48,14 @@
     "lefthook": "1.11.12",
     "lil-gui": "^0.20.0",
     "onchange": "7.1.0",
+    "spark-internal-rs": "file:rust/spark-internal-rs/pkg",
     "stats.js": "^0.17.0",
-    "three": "^0.172.0",
+    "three": "git+https://github.com/mrxz/three.js.git#array-texture-mrt-build",
     "ts-node": "10.9.2",
     "typescript": "^5.7.3",
     "vite": "^6.0.11",
     "vite-plugin-dts": "^4.5.4",
-    "vite-plugin-glsl": "^1.3.1",
-    "spark-internal-rs": "file:rust/spark-internal-rs/pkg"
+    "vite-plugin-glsl": "^1.3.1"
   },
   "dependencies": {
     "fflate": "^0.8.2"
--- a/src/PackedSplats.ts
+++ b/src/PackedSplats.ts
@@ -12,7 +12,7 @@ import {
   outputPackedSplat,
 } from "./dyno";
 import { TPackedSplats, definePackedSplats } from "./dyno/splats";
-import computeUvec4Template from "./shaders/computeUvec4.glsl";
+import computeUvec4x2Template from "./shaders/computeUvec4x2.glsl";
 import { getTextureSize, setPackedSplat, unpackSplat } from "./utils";
 
 // Initialize a PackedSplats collection from source data via
@@ -38,7 +38,7 @@ export type PackedSplatsOptions = {
   maxSplats?: number;
   // Use provided packed data array, where each 4 consecutive uint32 values
   // encode one "packed" Gsplat. (default: undefined)
-  packedArray?: Uint32Array;
+  packedArray?: [Uint32Array, Uint32Array];
   // Override number of splats in packed array to use only a subset.
   // (default: length of packed array / 4)
   numSplats?: number;
@@ -59,7 +59,7 @@ export type PackedSplatsOptions = {
 export class PackedSplats {
   maxSplats = 0;
   numSplats = 0;
-  packedArray: Uint32Array | null = null;
+  packedArray: [Uint32Array, Uint32Array] | null = null;
   extra: Record<string, unknown>;
 
   initialized: Promise<PackedSplats>;
@@ -68,7 +68,7 @@ export class PackedSplats {
   // Either target or source will be non-null, depending on whether the PackedSplats
   // is being used as a data source or generated to.
   target: THREE.WebGLArrayRenderTarget | null = null;
-  source: THREE.DataArrayTexture | null = null;
+  source: THREE.DataArrayTexture[] | null = null;
   // Set to true if source packedArray is updated to have it upload to GPU
   needsUpdate = true;
 
@@ -106,7 +106,7 @@ export class PackedSplats {
       // Calculate number of horizontal texture rows that could fit in array.
       // A properly initialized packedArray should already take into account the
       // width and height of the texture and be rounded up with padding.
-      this.maxSplats = Math.floor(this.packedArray.length / 4);
+      this.maxSplats = Math.floor(this.packedArray[0].length / 4);
       this.maxSplats =
         Math.floor(this.maxSplats / SPLAT_TEX_WIDTH) * SPLAT_TEX_WIDTH;
       this.numSplats = Math.min(
@@ -152,7 +152,8 @@ export class PackedSplats {
       this.target = null;
     }
     if (this.source) {
-      this.source.dispose();
+      this.source[0].dispose();
+      this.source[1].dispose();
       this.source = null;
     }
   }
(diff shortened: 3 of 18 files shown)


asundqui:
Wow, I was literally thinking of a similar thing myself! I have something else in mind for those 32 bytes though: I think we can get HDR emissive splats, PBR materials, and a continuous level-of-detail representation all in there at the same time. Let me follow up later with more on this, great thinking!

asundqui:
I actually also think we can do straight 2D textures instead of arrayed textures, if we're willing to accept a maximum # "active splats". For example, using 6x 4096^2 textures (which would be broadly supported across devices) can get us up to 96M active splats, which I think would be sufficient for anything that could conceivably render fast enough on a user device anyway.

asundqui:
@mrxz I have some previous experiencing implementing this for LoD: https://repo-sam.inria.fr/fungraph/hierarchical-3d-gaussians/

I think it's overkill however and I'm thinking of a simpler approach that will cover what we need for LoD and beyond. Here are the quantities and bit depths I'm thinking of storing that will enable us to do proper LoD but also PBR materials:
- Center x,y,z: float32 (12 bytes)
- Opacity float16 or Opacity8 + Shape8 to allow reshaping the splat profile up the LoD hierarchy (2 bytes)
- PR Metalness8, Roughness8 (2 bytes)
- RGBE8 (emissive color) / RGB8 (albedo) encoded together (4 bytes)
- Rotation OctXY101010 (expanded resolution from 888 in PackedSplats) (4 bytes)
- Scale X/Y/Z/min0/min/max/max0 in 9-bit log (8 bytes / 63 bits). XYZ scales for the splats. min0/min/max/max0 are four quantities that define the LoD scale levels: below min0 => opacity =0, min0..min => opacity fades to 1, min..max => opacity=1, max..max0 => opacity fades to 0

This adds up to 32 bytes nicely, gives us more resolution where it matters, and I think will pave the way for both HDR splats at the same time as PBR albedo/metal/rough properties, and a very general approach to being able to handle continuous LoD levels gracefully.

I think moving to 32-bytes per splat, even without texture array, would make sense as an option. It will impact performance slightly from being more memory intensive, but I think it can be a no-compromises solution for everything else we want.

asundqui:
@mrxz went through your changes, they look great, however I'd like to do something more along what I'm describing above: instead of changing PackedSplats, I was thinking of introducing a second class "ExtSplats" or similar, explicitly differentiating PackedSplats from ExtSplats. This will allow the user to decide which representation they want depending on their case. Your draft here going through all the code spots that need to be changed to accommodate this is very helpful though! Are you okay if I follow up and build on your work here in a separate branch? I also am planning on introducing extended encoding ranges as part of PackedSplats as well.

mrxz:
> I actually also think we can do straight 2D textures instead of arrayed textures, if we're willing to accept a maximum # "active splats". For example, using 6x 4096^2 textures (which would be broadly supported across devices) can get us up to 96M active splats, which I think would be sufficient for anything that could conceivably render fast enough on a user device anyway.

The drawback with straight 2D textures is that WebGL2 requires "constant-index-expressions" if we were to present these textures as an array of samplers. So you'd either have multiple draw calls or not-so-nice conditionals in the shader. Though the argument can be made that devices struggling with this would probably not fare well with > 16M splats either way.

If we do stick with `TEXTURE_2D_ARRAY`, support for it has been merged into Three.js and should be available from r179 onwards: https://github.com/mrdoob/three.js/pull/31476

> This adds up to 32 bytes nicely, gives us more resolution where it matters, and I think will pave the way for both HDR splats at the same time as PBR albedo/metal/rough properties, and a very general approach to being able to handle continuous LoD levels gracefully.

Looks good. The additional resolution is most needed for the center IMHO, as users can easily run into issues when loading an off center splat file or positing things away from the origin.

As for the PBR and HDR properties, I find it hard to say how this will end up working. Conceptually the PBR properties make sense for surfaces, which 2D gaussians could approximate. But with the blending taking place, I don't see how the result would be energy conserving.

But it's clear that 32 bytes per splat gives us ample space to both increase the resolution for the properties that need it and have room to spare for future features.

> Your draft here going through all the code spots that need to be changed to accommodate this is very helpful though! Are you okay if I follow up and build on your work here in a separate branch? I also am planning on introducing extended encoding ranges as part of PackedSplats as well.

Sure, no problem. Note that in this draft PR I hadn't updated the raycasting yet, which obviously will have to be able to unpack the relevant splat properties as well.

Looking forward to see how things turn out. Given the user the option between `PackedSplat` and `ExtSplats` is nice, though I'm slightly concerned about the additional complexity this might bring. Though at the very least the loading of splats abstracts the packing away quite nicely, which should making alternate representations relatively easy to support.

asundqui:
> The drawback with straight 2D textures is that WebGL2 requires "constant-index-expressions" if we were to present these textures as an array of samplers. So you'd either have multiple draw calls or not-so-nice conditionals in the shader. Though the argument can be made that devices struggling with this would probably not fare well with > 16M splats either way.
> 
> If we do stick with `TEXTURE_2D_ARRAY`, support for it has been merged into Three.js and should be available from r179 onwards: [mrdoob/three.js#31476](https://github.com/mrdoob/three.js/pull/31476)

That's frickin AWESOME! I can't believe you got it into Three.js so fast, that unlocks so much potential for Spark!! It's been killing me to not be able to render more than one uvec4 array target at a time. Any concerns about forcing users to upgrade to r179? I guess we just have to communicate it, maybe I throw a warning/error if the version is older?

> Looks good. The additional resolution is most needed for the center IMHO, as users can easily run into issues when loading an off center splat file or positing things away from the origin.

Agreed, float32 there makes everything so much simpler and better.

> As for the PBR and HDR properties, I find it hard to say how this will end up working. Conceptually the PBR properties make sense for surfaces, which 2D gaussians could approximate. But with the blending taking place, I don't see how the result would be energy conserving.
> But it's clear that 32 bytes per splat gives us ample space to both increase the resolution for the properties that need it and have room to spare for future features.

Yeah, it may not be energy conserving, but my gut tells me it will actually look pretty good and correct! I need to experiment with migrating splatVertex/splatFragment.glsl to TSL, which I think could allow us to just "slot in" with Three.js's standard material shader and lighting/shadowing etc. That's my hope, I haven't had time to experiment with it yet.

In any case, as you said the 32 bytes gives us plenty of room to experiment with features like this, and we can change the implementation details over time as we see fit.

> > Your draft here going through all the code spots that need to be changed to accommodate this is very helpful though! Are you okay if I follow up and build on your work here in a separate branch? I also am planning on introducing extended encoding ranges as part of PackedSplats as well.
> 
> Sure, no problem. Note that in this draft PR I hadn't updated the raycasting yet, which obviously will have to be able to unpack the relevant splat properties as well.
> 
> Looking forward to see how things turn out. Given the user the option between `PackedSplat` and `ExtSplats` is nice, though I'm slightly concerned about the additional complexity this might bring. Though at the very least the loading of splats abstracts the packing away quite nicely, which should making alternate representations relatively easy to support.

Yes, I share a little concern there as well, but my hope is that these two will be sufficient for 99.9% of use cases? PackedSplats is arguably sufficient for more use cases I'd argue... I guess if one day there is a 3rd type we want to introduce, we could factor out some base thing with an interface...

Okay let me fiddle with this thing, maybe I'll put out a PR with a smaller set of encoded values first.

mrxz:
> That's frickin AWESOME! I can't believe you got it into Three.js so fast, that unlocks so much potential for Spark!! It's been killing me to not be able to render more than one uvec4 array target at a time. Any concerns about forcing users to upgrade to r179? I guess we just have to communicate it, maybe I throw a warning/error if the version is older?

The main concern would be that it might limit who can use Spark. For new projects it shouldn't be a problem to use the latest Three.js version. For existing projects or projects indirectly using Three.js through a framework or engine, they might be "stuck" on an older version.

Either way, it would be a good idea to set the minimum required Three.js version as `peerDependency` in the `package.json`. That way the package manager of the user should warn them when the requirement can't be met.

> Yeah, it may not be energy conserving, but my gut tells me it will actually look pretty good and correct! I need to experiment with migrating splatVertex/splatFragment.glsl to TSL, which I think could allow us to just "slot in" with Three.js's standard material shader and lighting/shadowing etc. That's my hope, I haven't had time to experiment with it yet.

TSL only works with the `WebGPURenderer` and not `WebGLRenderer` (see https://github.com/mrdoob/three.js/issues/30185). Ultimately Three.js intends to deprecate the `WebGLRenderer` so we'd have to use TSL or otherwise ensure it works with both the WebGL2 and WebGPU backends. Contrary to its name `WebGPURenderer` has both a WebGL2 and WebGPU backend.

Haven't experimented a lot with `WebGPURenderer` yet, as it didn't support WebXR for the longest time. Now there is some support when using the WebGL2 backend. I'm concerned the transition will put a lot of Three.js libraries at a predicament: focus entirely on WebGPU or effectively maintain two versions. Unlike forcing users to a specific Three.js version, forcing them into a specific renderer is probably more limiting.

jo-chemla:
Has this work been replaced by another tweak to handle 32 bytes splats properties representation within sparkjs? Or is this no more required given how you encode and pack position, rotation etc using other underlying format or container spec?

mrxz:
> Has this work been replaced by another tweak to handle 32 bytes splats properties representation within sparkjs? Or is this no more required given how you encode and pack position, rotation etc using other underlying format or container spec?

Spark 2.0 has introduced support for `ExtSplat`, see: https://sparkjs.dev/2.0.0-preview/docs/ext-splats/#32-byte-layout

This representation can be enabled per `SplatMesh` and on the `SplatAccumulator` and `SplatPager` using the `extSplats`, `accumExtSplats` and `pagedExtSplats` properties respectively. For most use cases this should give enough precision.

jo-chemla:
Thanks for getting back so fast I had missed that, nice!

jo-chemla:
Edit: just submitted a feature request for exposing a pure js api RAD hlod tree walker, which could help convert the sparkjs tiled rad/tadc hierarchy into OGC 3Dtiles with KHR_gaussian_splatting extension glb. 
https://github.com/sparkjsdev/spark/issues/372

PR #129, opened 2025-07-21
Implement float32 sorting option

asundqui:
This PR adds an option `sort32` to `SparkViewpoint` to enable float32 sorting precision, implemented via 2-pass radix sort (base 65536) on CPU worker. Each SparkViewpoint can choose for itself whether to use float32 precision or the default float16 mode, so you could have a "faster" viewpoint for real-time display, and a "more accurate" viewpoint that's doing a video render-out using higher precision.

The below video shows an example of improvement possible with this, albeit with settings designed to "exaggerate" high-contrast changes with non-AA and low-DPI:

https://github.com/user-attachments/assets/09171e6f-d28e-4b34-bbac-ed61450361c4

Benchmarks show the new float32 sort is 2.5-3x slower than float16 sort, so giving the user an option gives us the best of both worlds: high-precision for users that care, higher performance if you don't care:

| Browser | JS16 (ms) | Wasm16 | JS32 | Wasm32 |
|---|---|---|---|---|
| Safari | 1.704 | 1.430 | 4.677 | 3.934 |
| Chrome | 3.242 | 1.411 | 7.189 | 3.808 |
| Firefox | 1.563 | 1.558 | 5.502 | 4.283 |

Added option to examples/editor to select the mode, default to float32 on.

### DIFF
--- a/docs/docs/spark-viewpoint.md
+++ b/docs/docs/spark-viewpoint.md
@@ -23,6 +23,7 @@ const viewpoint = spark.newViewpoint({
   sortCoorient?: boolean;
   depthBias?: number;
   sort360?: boolean;
+  sort32?: boolean;
 });
 ```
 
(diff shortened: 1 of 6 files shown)


61cygni:
This is fantastic. Thank you!

mrxz:
The tallying in `bucket16` for the second pass does not depend on the output of the first pass. So separate arrays would allow them both to be tallied at the same time: see #132

PR #130, opened 2025-07-21
Externalize render settings minAlpha and maxPixelRadius

asundqui:
Built on top of https://github.com/sparkjsdev/spark/pull/129

This PR externalizes two render settings that were previously hard-coded:
- minAlpha: minimum alpha value to render anything. Currently 0.5/255 in order to render the lowest-opacity of 1/255 with an allowance for rounding. Most scans/3DGS are not so sensitive to low-opacity splats and can often be discarded up to some minimum level (for example minAlpha=0.1 is likely to result in identical output)
- maxPixelRadius: splats that get too large are clamped to a maximum radius, currently set to 512. Generally only visible effect when you are "very close" to splats such that they begin to fill up the screen. Setting this lower can improve performance, while setting higher will result in more faithful rendering.

We may also want to add a `minPixelRadius` or similar. PlayCanvas will discard splats that are < 2 pixels in size, and perhaps we should also do the same for performance reasons. Doing this unconditionally is probably a bad idea but perhaps adding a setting would allow the user to choose performance vs. fidelity here as well.

We should also consider externalizing any other render settings/parameters that the user might want to adjust. I think it's a good thing to give the user as much control as possible!


dmarcos:
Awesome thanks!

PR #131, opened 2025-07-22
Added particle flow transition to examples

kali-shade:
Splat Flow Effect

This effect animates a smooth "flow" of 3D splats from one object to another. Instead of instant teleportation, the splats appear to dematerialize from the center of the bounding box of the current object and rematerialize at the center of the next. The transition uses bounding box center calculation, dynamic scaling, opacity modulation, and wave-based distortion to create a visually continuous migration of particles between shapes.

### DIFF
--- a/examples/splat-flow/index.html
+++ b/examples/splat-flow/index.html
@@ -1,0 +1,258 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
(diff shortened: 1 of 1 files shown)


asundqui (approved):
Wonderful, thanks for contributing!

@dmarcos should he add it to `examples.html`, or `index.html`, or is it both??

dmarcos (approved):

dmarcos:
> Wonderful, thanks for contributing!
> 
> @dmarcos should he add it to `examples.html`, or `index.html`, or is it both??

Yeah. I’ll add it

PR #134, opened 2025-07-22
Custom/extended splat encoding ranges, Premultiplied alpha for hyper-saturated colors

asundqui:
Added the `SplatEncoding` concept (defined in `PackedSplats.ts`) that stores user-configurable ranges for RGB values (rgbMin/rgbMax), scale X/Y/Z (lnScaleMin/Max), as well as sh1Min/Max, sh2, sh3.

The default values are always rgbMin/Max=(0,1), sh?Min/Max=(-1,1), lnScaleMin/Max=(-12,9), and should be fully backward-compatible with the current Spark behavior. You can adjust these settings on both `SparkRenderer` (which changes the allowed encoding ranges in the `SplatAccumulator` that aggregates all scene splats), and on `PackedSplats` (within a `SplatMesh`), which indicates the encoding used for the stored splats.

It's advisable to use the same ranges for both SparkRenderer and your SplatMeshes because otherwise you may incur double-rounding error.

This PR also adds a (default on) `SparkRenderer.premultipliedAlpha` mode, which adjusts the blend mode allowing us to blend hyper-saturated colors (when RGB > A). Without this setting the encoding ranges have no effect because RGB values are always clamped to 1.

Examples/editor has been updated with new settings under the "Debug" menu where you can change the values. Changing SplatMesh encoding requires loading a new file for the new ranges to take effect.

I believe these new settings allow the user to fix the contrast issue in #81 , see the video below:

https://github.com/user-attachments/assets/f8682115-17e6-4a90-8e5d-f7a59b393003

There's a substantial amount of piping needed to send all the encodings where it needs to go (including Rust), I hope I've covered everything!

In the future when we add an "ExtendedSplats" sort of concept, we can reuse this infra hopefully to allow configurable scaling there as well.

### DIFF
--- a/examples/editor/index.html
+++ b/examples/editor/index.html
@@ -69,7 +69,7 @@
     import * as THREE from "three";
     import { OrbitControls } from "three/addons/controls/OrbitControls.js";
     import { GUI } from "lil-gui";
-    import { constructGrid, SparkControls, SparkRenderer, SplatMesh, textSplats, dyno, transcodeSpz, isMobile, isPcSogs } from "@sparkjsdev/spark";
+    import { constructGrid, SparkControls, SparkRenderer, SplatMesh, textSplats, dyno, transcodeSpz, isMobile, isPcSogs, LN_SCALE_MIN, LN_SCALE_MAX } from "@sparkjsdev/spark";
     import { getAssetFileURL } from "/examples/js/get-asset-url.js";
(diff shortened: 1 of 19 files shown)


61cygni:
Looks great. I cloned and run through all the examples and spot checks seem OK. Thank you!

PR #135, opened 2025-07-23
Stochastic sort-free splat rendering (Experimental feature)

asundqui:
This PR adds an experimental rendering mode via `SparkViewpoint.stochastic` (default false). By setting it to true, the rendering mode is switched to fully-opaque splats that are rendered/discarded stochastically with probability equal to the blending alpha. Since the Z-buffer manages fully opaque depth ordering we get "stochastically correct" ordering of the splats without needing to sort them to draw them in a particular order. In my testing I have observed 2x+ rendering frame rate improvement from this change.... My theory is that it mostly comes from being able to traverse all the splat properties in linear/consecutive order rather than sampling randomly according to depth sorting.

https://github.com/user-attachments/assets/82506ce6-bca4-42a9-88dd-d852825f5631

Loosely inspired by this work: https://arxiv.org/pdf/2503.24366
In that work they take multiple samples and average. In this case I simply "render fast" and let the averaging happen over time :)

This PR diff is confusing because it is stacked on an older commit main and has a lot of unrelated SplatEncoding stuff. Only the last three commits "Checkpoint" and down are relevant to this, and the experimental rendering mode is gated behind a "stochastic" flag. 

https://github.com/user-attachments/assets/5cf83324-46e3-4753-a2c3-44ad655565e9

Anyway I think stochastic splat rendering was mentioned at some point in our Discord, this PR lets us experiment with this interesting idea!

### DIFF
--- a/examples.html
+++ b/examples.html
@@ -241,6 +241,7 @@
           'debug-color': '/examples/debug-color/index.html',
           'depth-of-field': '/examples/depth-of-field/index.html',
           'splat-texture': '/examples/splat-texture/index.html',
+          'stochastic': '/examples/stochastic/index.html',
           'editor': '/examples/editor/index.html'
         };
 
@@ -440,6 +441,7 @@
           <a href="#debug-color" data-example="debug-color" class="example-link">Debug Coloring</a>
           <a href="#depth-of-field" data-example="depth-of-field" class="example-link">Depth of Field</a>
           <a href="#splat-texture" data-example="splat-texture" class="example-link">Splat Texture</a>
+          <a href="#stochastic" data-example="stochastic" class="example-link">Stochastic Rendering</a>
           <a href="#editor" data-example="editor" class="example-link">Editor</a>
         </div>
       </div>
--- a/examples/stochastic/index.html
+++ b/examples/stochastic/index.html
@@ -1,0 +1,78 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8">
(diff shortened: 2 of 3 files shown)


mrxz:
Really cool. 

> My theory is that it mostly comes from being able to traverse all the splat properties in linear/consecutive order rather than sampling randomly according to depth sorting.

My guess would be that it's mostly due to the depth testing allowing a lot of fragment shading to be skipped. While the fragment shader itself isn't overly expensive, the required back-to-front sorting for blending is literally the worst case scenario.

While the linear order is definitely more cache friendly, I don't know if the GPU is going to perform texture prefetching in the vertex shader for a texture indexed using instance ID. In which case the gains would be limited. I also seem to recall that the GPU metrics on the Quest 3 didn't indicate being stalled on texture fetching, though I haven't profiled with this change and could be misremembering things.

> Anyway I think stochastic splat rendering was mentioned at some point in our Discord, this PR lets us experiment with this interesting idea!

The interesting part will be to find a way to have more samples, possible using something like TAA. Though getting accurate motion vectors might be difficult. And there's the real risk that introducing the motion vector and history buffers is going to negate a lot of the performance gains on mobile GPUs.

asundqui:
Yeah, it's possible a 1-frame TAA thing could bring this into the realm of comparable with regular blending, one day someone can experiment with that :) Likely could undo the performance gains. Would love to see in the future if someone thinks of a use case for this where not having to sort makes all the difference and this actually is beneficial :).

PR #136, opened 2025-07-24
Use `material.premultipliedAlpha` to adjust blending function

mrxz:
With #134 the default blending now uses `premultipliedAlpha`. The material system of Three.js has built-in support for this, which takes care of setting the right blending function as well as providing a define in the fragment shader.

This PR replaces the custom blending function and instead uses the `premultipliedAlpha` setup. Additionally the uniform has been replaced with the define. This does mean that changing the flag requires a shader recompilation, though these variants are cached by Three.js. In the unlikely situation that a user repeatedly switches this flag, it should only incur a recompilation once.

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -294,14 +294,8 @@ export class SparkRenderer extends THREE.Mesh {
       vertexShader: shaders.splatVertex,
       fragmentShader: shaders.splatFragment,
       uniforms,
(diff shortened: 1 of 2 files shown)


asundqui (approved):
Nice! Very helpful with in-depth Three.js knowledge, the trade-off sounds good to me, unlikely the user will be changing that flag I agree, but even so seems fine.

Issue #145, opened 2025-08-01
Alleviate of rendering appearing noisy due to sh3 precision limitations


heimeii:
https://github.com/sparkjsdev/spark/issues/142#issuecomment-3139361018

encode sh3 to sint6 min precision is 1/31 (0.03225)

if use uin8 replace, min precision is 1/127 (0.00787)

Based on my current observations, the value of sh3 frequently tends to be small, which is highly likely the cause of this phenomenon.

in sh3 buffer, item total size is 128 and used 126, 2bit is free

so, we can real calculate scale to alleviate precision.

code look like (only test code):

<img width="1128" height="980" alt="Image" src="https://github.com/user-attachments/assets/b64783fd-0b3f-44dc-9aef-a60166ce703d" />

<img width="905" height="373" alt="Image" src="https://github.com/user-attachments/assets/e2401be9-17fc-44e2-b90b-6b56052af2ee" />

before:

<img width="840" height="783" alt="Image" src="https://github.com/user-attachments/assets/ff9fa8c9-b2af-46bf-8f20-af6697dc8b50" />

after:

<img width="737" height="713" alt="Image" src="https://github.com/user-attachments/assets/c21ebf58-df9e-4882-8036-702322195b5c" />

Although there has been some improvement, it still falls short of supersplat level of cleanliness

heimeii:
can use 2 ** flags

heimeii:
after opt parse perf in wasm is 1.2s, spark now is 2.3

<img width="1000" height="262" alt="Image" src="https://github.com/user-attachments/assets/788f1f87-195b-4ddb-abaa-af42648c0c1b" />

mrxz:
Could you try with #146? It increases the SH3 coefficients from `sint6` to `sint8` which should allow for higher precision.

heimeii:
> Could you try with [#146](https://github.com/sparkjsdev/spark/pull/146)? It increases the SH3 coefficients from `sint6` to `sint8` which should allow for higher precision.

Thank you for the implementation suggestions, though they might not be suitable for my scenario. I suspect they could cause performance degradation on low-end devices, as GPU texture fetch caching is relatively unfriendly to linear memory access patterns. I'll need to wait until Monday to test this empirically.

Issue #142, opened 2025-07-31
Looks dirty and with some artifacts compared to Supersplat


heimeii:
spark (no-aa: 0.3, aa: 0, focalAdjustment: 2, premul alpha)

<img width="1920" height="908" alt="Image" src="https://github.com/user-attachments/assets/84894a0a-d5e3-4d14-93f4-55f1c076f28f" />

supersplat

<img width="1543" height="705" alt="Image" src="https://github.com/user-attachments/assets/7ee24e81-67d9-4885-a2e0-e0459129ac08" />

Supersplat looks cleaner on the wall, which seems to be related to the color precision.

You can get the test file from
https://drive.google.com/file/d/1Ywt6RJ_OmBJCSReqJt5er9nPS-AVaBa_/view?usp=drive_link

Regards.

heimeii:
It seems to be related to this change https://github.com/sparkjsdev/spark/pull/134

I think the calculation of shn scale doesn't seem correct.

sh1 scale 63  -> 126
sh2 scale 0.5 -> 2
sh3 scale 31 -> 64

<img width="1332" height="587" alt="Image" src="https://github.com/user-attachments/assets/fac2b386-5855-4927-b133-0d6c1ebdbaa7" />

current:

<img width="1920" height="897" alt="Image" src="https://github.com/user-attachments/assets/8ebe4bb7-c219-4ea1-b20c-eae5118380e5" />

fixed:

<img width="1920" height="910" alt="Image" src="https://github.com/user-attachments/assets/dee3dc0b-cb44-4341-b742-b32ccac2641a" />

But it still looks a bit dirty

mrxz:
> It seems to be related to this change [#134](https://github.com/sparkjsdev/spark/pull/134)
> 
> I think the calculation of shn scale doesn't seem correct.
> 
> sh1 scale 63 -> 126
> sh2 scale 0.5 -> 2
> sh3 scale 31 -> 64

@heimeii Good catch, can confirm that the issue has to do with the spherical harmonics. For the sh3 scale it should be 62. If you want, feel free to open a PR with these changes.

With the above fixes the difference is indeed smaller, though some discolouration can still be seen with Spark:
![Image](https://github.com/user-attachments/assets/b888c4c0-181f-432f-ac1e-555053ed9b73)

This only seems to be an issue with the third degree spherical harmonics, as limiting it to 2 (`maxSh = 2`) gives the following:
[...]

PR #146, opened 2025-08-01
Interleave spherical harmonics coefficients

mrxz:
This PR changes the way spherical harmonics are stored in `PackedSplats.extra` and how they are evaluated. Instead of keeping the coefficients for the different orders separate, they are interleaved in a single buffer. Their representation is unified to be `sint8`.

Main motivation was to address the remaining quality issue reported in #145 (see comparison: https://github.com/sparkjsdev/spark/issues/142#issuecomment-3139361018). The current representations (`sint7`, `sint8` and `sint6`) were chosen such that they fit in a single texel. Simply upping this to `sint8` would require multiple texture fetches _per_ order and introduces a lot of padding. By interleaving them the padding is minimized, keeping the amount of texture fetches required low.

| SH Degrees | Coefficients (cumulative) | Texels / Format | Padding | 
|--|--|--|--|
|1 | 9 | `RGB` (3 * 4 = 12) | 3 |
|2 | 24 (9 + 15) | 2 x `RGB` (2 * 3 * 4 = 24) | 0 |
|3 | 45 (9 + 15 + 21) | 3 x `RGBA` (3 * 4 * 4 = 48) | 3 |

With the above there's only a single `sh` texture. Since the exact layout of the buffer and the resulting texture depends on the number of degrees in the source file, the `maxSh` property of the `SplatMesh` only changes which orders are evaluated, not how they are read. This means multiple combinations exist, which are handled using `#define/#if` directives in the GLSL code.

There are a couple of upsides:

- Improves the visual quality (#145)
- Reduces time spent during loading/parsing as less work is done (observed with the SOGS example)
- All SH coefficients are closely packed in the buffer and textures

Additionally I've added a `maxSh` property to `SplatEncoding`. This allows someone to skip loading and encoding the sh coefficients for orders they don't intend on using. While not fully utilized in all loaders, this can further improve loading/parsing speeds as well as save memory usage.

I haven't done extensive performance testing. In particular the RGB texture format might perform poorly on certain hardware, but it does allow for the least amount of padding. Not sure if in those cases having 3x`RG` for `maxSh = 2` would perform better or worse. But before diving into that, let's first establish if this approach is the way to go.

### DIFF
--- a/src/PackedSplats.ts
+++ b/src/PackedSplats.ts
@@ -33,6 +33,7 @@ export type SplatEncoding = {
   sh2Max?: number;
   sh3Min?: number;
   sh3Max?: number;
+  maxSh?: number;
 };
 
 export const DEFAULT_SPLAT_ENCODING: SplatEncoding = {
@@ -46,6 +47,7 @@ export const DEFAULT_SPLAT_ENCODING: SplatEncoding = {
   sh2Max: 1,
   sh3Min: -1,
   sh3Max: 1,
+  maxSh: 3,
 };
 
 // Initialize a PackedSplats collection from source data via
--- a/src/SplatMesh.ts
+++ b/src/SplatMesh.ts
@@ -38,7 +38,7 @@ import {
   unindent,
   unindentLines,
 } from "./dyno";
-import { getTextureSize } from "./utils";
+import { getShArrayStride, getTextureSize } from "./utils";
 
 export type SplatMeshOptions = {
   // URL to fetch a Gaussian splat file from(supports .ply, .splat, .ksplat,
@@ -380,9 +380,8 @@ export class SplatMesh extends SplatGenerator {
 
         if (this.maxSh >= 1) {
           // Inject lighting from SH1..SH3
-          const { sh1Texture, sh2Texture, sh3Texture } =
-            this.ensureShTextures();
-          if (sh1Texture) {
+          const { shTexture, shDegrees } = this.ensureShTexture();
+          if (shTexture && shDegrees) {
             //Calculate view direction in object space
             const viewCenterInObject = viewToObject.translate;
             const { center } = splitGsplat(gsplat).outputs;
@@ -399,22 +398,16 @@ export class SplatMesh extends SplatGenerator {
             }
 
             // Evaluate Spherical Harmonics
-            const sh1Snorm = evaluateSH1(gsplat, sh1Texture, viewDir);
-            let rgb = rescaleSh(sh1Snorm, this.packedSplats.dynoSh1MinMax);
-            if (this.maxSh >= 2 && sh2Texture) {
-              const sh2Snorm = evaluateSH2(gsplat, sh2Texture, viewDir);
-              rgb = add(
-                rgb,
-                rescaleSh(sh2Snorm, this.packedSplats.dynoSh2MinMax),
-              );
-            }
-            if (this.maxSh >= 3 && sh3Texture) {
-              const sh3Snorm = evaluateSH3(gsplat, sh3Texture, viewDir);
-              rgb = add(
-                rgb,
-                rescaleSh(sh3Snorm, this.packedSplats.dynoSh3MinMax),
-              );
-            }
+            const rgb = evaluateSH(
+              gsplat,
+              shTexture,
+              viewDir,
+              shDegrees,
+              this.maxSh,
+              this.packedSplats.dynoSh1MinMax,
+              this.packedSplats.dynoSh2MinMax,
+              this.packedSplats.dynoSh3MinMax,
+            );
 
             // Flash off for 0.3 / 1.0 sec for debugging
             // const fractTime = fract(SplatMesh.dynoTime);
@@ -520,7 +513,7 @@ export class SplatMesh extends SplatGenerator {
     const viewToObjectMatrix = worldToObject.multiply(viewToWorld);
     if (
       viewToObject.updateFromMatrix(viewToObjectMatrix) &&
-      (this.enableViewToObject || this.packedSplats.extra.sh1)
+      (this.enableViewToObject || this.packedSplats.extra.sh)
     ) {
       // Only trigger update if we have view-dependent spherical harmonics
       updated = true;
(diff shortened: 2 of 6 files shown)


dmarcos:
Thanks. This is awesome. Happy to see that precision up!

mrxz:
Did some profiling of the draw call that evaluates the spherical harmonics using [`EXT_disjoint_timer_query_webgl2`](https://registry.khronos.org/webgl/extensions/EXT_disjoint_timer_query_webgl2/). On desktop the difference was negligible, but the impact on an Android smartphone and Quest 3 was considerably larger (> 40%). 

Changing the representation to separate textures sampled at the same location had a considerably smaller impact. So with 3 `RGBA32UI` textures holding the 45 coefficients, the impact was ~8% on the smartphone and ~10% on the Quest 3. 

Interestingly enough the time it saves while loading/parsing splat files is more than it takes to split the coefficients per texture. Going to rework this PR to support both old and new mode (specified through `splatEncoding`) and try to retain the parsing speed gains.

asundqui:
@mrxz I like the general idea of increasing the precision of the SH coefficients to be uniformly 8 bits, and it seems from #145 and from some other comments we saw on our Discord on SPZ file format's bit reduction that we get some banding as a result. I hope that 8 bits would be enough, but I'm worried that even that wouldn't do it. I don't think we can afford float16 from a memory/performance perspective... but should we allocate more than 8 bits possibly? Of maybe we need some off-line tool for people to analyze their PLY and get recommendations for shNmin/max?

Not entirely a serious idea, but I wonder if a "random dithered rounding" could make the banding basically imperceptible, or if it would just look noisy :). So basically when rounding to sint8 (or lower bits) we add (rand() - 0.5) to the values and then round those.

My impression with RGB-style textures is that they are generally discouraged because some hardware isn't equipped to deal with data stride 3 efficiently. I wonder if that's what was going on with Quest 3 / Android? Another theory I have is that the pseudo-compute shader presumably executes over "tiles", and if those tile sizes happen to match the tile size of the texture lookups, then it's almost like we are guaranteed good texture cache efficiency. But if the texture size doesn't match then the texture access pattern gets spread over multiple tiles in a strange way. This was the mental model I had when I originally coded up the sh1-3 textures.

The most annoying thing is SH1 with 9 coefficients... 1 more than a power of 2! I can't think of a great way to pack things such that we have 8 bits per coefficient and don't waste too much memory (and bandwidth) :/ I think if we can be okay with `maxSh=1` being very inefficient (likely much less common than `maxSh>=2`), and just used RGBA32UI for all textures instead of RG32UI for SH1? Then we can support up to SH3 with Sint8 coefficients like you described in your testing today.

If you can make the PR support both old and new modes then that would be great from a user's perspective - they can choose between slightly more memory efficient, or higher bit precision. Should we be worried about the code complexity from all these options?? I guess there are other places in Spark that are getting pretty complex :)

Anyway, great work, I support this line of inquiry, as long as there are options so that Android/Quest users won't be impacted with no recourse!

mrxz:
> I hope that 8 bits would be enough, but I'm worried that even that wouldn't do it. I don't think we can afford float16 from a memory/performance perspective... but should we allocate more than 8 bits possibly? Of maybe we need some off-line tool for people to analyze their PLY and get recommendations for shNmin/max?

I would assume 8 bits should suffice given how many of the formats seem to use that as well (SOGS, Playcanvas Compressed .ply, SPZ). We should try to make the most out of these 8 bits, as there are two critical points:

- The loaders parse the coefficients as floating point numbers which are then encoded again. Mismatch in shNmin/max can cause loss of precision
- When evaluating the spherical harmonics the end result is encoded into the RGB values. This can introduce clamping and loss of precision.

Instead of an offline tool, I've been wondering if maybe there should be a mechanism for the loaders to suggest suitable `SplatEncoding` values. Obviously there's a bit of a chicken-and-egg problem here as the loaders also perform the encoding. And it might be difficult to find a clean way to propagate that encoding to the `SparkRenderer`.

But the current `SplatEncoding` seems to provide a technical solution, but not a practical one for many users. An offline tool could help, but it'll only be a matter of time until people would want to run it online (as part of their app/experience).

> My impression with RGB-style textures is that they are generally discouraged because some hardware isn't equipped to deal with data stride 3 efficiently. I wonder if that's what was going on with Quest 3 / Android?

That's what I understand to be the main reason as well, though I've also heard that this only applies to 8-bit channels. Either way I prefer trying things instead of just assuming, as there are many outdated best-practices floating around. And it never hurts to get an idea/quantify the impact it has.

> Another theory I have is that the pseudo-compute shader presumably executes over "tiles", and if those tile sizes happen to match the tile size of the texture lookups, then it's almost like we are guaranteed good texture cache efficiency. But if the texture size doesn't match then the texture access pattern gets spread over multiple tiles in a strange way. This was the mental model I had when I originally coded up the sh1-3 textures.

The GPU in the Quest supports two modes: binned and direct. For the fullscreen passes it uses the direct mode, so they aren't binned into tiles. Though the fragment shader is of course still processed in (at least) 2x2 blocks. Clearly the separate textures perform better, whether due to caching or a more predictable access pattern for pre-fetching or something else.

It's possible that laying out the pixels in a different way in a single texture might yield similar performance, but at that point there wouldn't really be any benefit.

> The most annoying thing is SH1 with 9 coefficients... 1 more than a power of 2! I can't think of a great way to pack things such that we have 8 bits per coefficient and don't waste too much memory (and bandwidth) :/ I think if we can be okay with maxSh=1 being very inefficient (likely much less common than maxSh>=2), and just used RGBA32UI for all textures instead of RG32UI for SH1? Then we can support up to SH3 with Sint8 coefficients like you described in your testing today.

Indeed, it's a lot more straightforward to simply use RGBA32UI textures throughout. And while it's suboptimal for the `maxSh=1` case, it doesn't seem very common. And if it does become common/requested we can always consider adding a different/dedicated approach for it.

> If you can make the PR support both old and new modes then that would be great from a user's perspective - they can choose between slightly more memory efficient, or higher bit precision. Should we be worried about the code complexity from all these options?? I guess there are other places in Spark that are getting pretty complex :)

There are definitely some worries about the increasing complexity the more options and configurations we add. It does increase the cost of maintenance, makes it harder for users to know what they should use and when investigating user-reported issues there are more variables/unknowns, etc....
So it's a trade-off, as long as we avoid a combinatorial explosion. In this case the spherical harmonics evaluation is pretty self-contained within `SplatMesh` so it should be okay IMO.

mrxz:
Did some more testing/profiling with a singular texture. Instead of using `splatIndex * 3 + i`, computing the coordinate once and scaling the x component by 3. So using a texture of `(width*3, height)`, and offsetting using `texelFetchOffset`. The performance of this approach was better than the 3 separate textures on PC and my smartphone. However, testing on an older smartphone and Quest 3 showed mixed results.

Both of these devices give highly varying times from run to run. At one point the Quest 3 even reported a time that was 50% of the baseline. Trying to figure out what is causing this behaviour, I believe that the (async) Readback ends up getting capture in the timer query or otherwise impacting it. 

Looking at the timeline there seems to be a suboptimal pattern arising for these two devices. Since they can't reach the target frame-rate they'll have pending sort operations. Once the worker is done sorting it immediately kicks of the pending sort, which starts up a GPU workload (Readback). This happens right before the rAF and carries over into it, to the point that it delays the render work for the frame.

Ideally the pending sort should only kick-off _after_ the rAF. Now it naively attempts two sorts in between rAF callbacks. I don't think there's anyway to cleanly estimate if there's time for it, so heuristically allowing only one seems like a good middle-ground. It would also give the rAF code the opportunity to replace the pending sort with a more up-to-date one.
(Additionally if we'd expect the second one to complete before the next rAF, there'd be no need to actually sort the first readback)

This pattern seems to converge pretty quickly, which seems to be (partially) caused by the 4ms polling rate of `readRenderTargetPixelsAsync`.

Either way, going to try to see if deferring the pending sort to after rAF helps, as well as lowering the polling interval. Hopefully that concentrates the GPU workload right after the rAF and not right before with a period of idle. At the very least I hope this makes the measurements more consistent/reliable, though I could also try smaller splats for testing.

PR #147, opened 2025-08-01
Reduce small memory allocations when loading ply files

mrxz:
Did a quick profile of loading a (compressed) .ply file with 1M splats. The loading showed quite a few minor GC interruptions, so I looked where memory allocations took place that could be avoided. This lead to two main culprits:

1. Some of the utils instantiate `THREE.Quaternion` or `THREE.Vector3` objects. Using a singular instances that is re-used between invocations avoids this.
2. Some of the inner-loops of the ply loading used for-of loops, which sadly cause allocations. These have been replaced with simple for loops

The parsing went from ~3.84s to ~3.54s, while the JavaScript Allocations went down significantly. The Firefox Profile showed ~30MiB from just the for-of loops alone, which is now entirely avoided.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -499,6 +499,10 @@ export function setPackedSplatScales(
     (packedSplats[i4 + 3] & 0xff000000);
 }
 
+// Temporary storage used in `encodeQuatOCtXy88R8` and `decodeQuatOctXy88R8` to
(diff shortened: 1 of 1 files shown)


dmarcos:
Nice!

asundqui:
Amazing!

I had no idea that for-of loops caused memory allocations, thank you for doing the profiling!

PR #149, opened 2025-08-03
Splat reveal effects (weather)

kali-shade:
General Description
Interactive weather effects demo showcasing tornado and rain animations applied to 3D splat models. Features a dropdown interface to switch between effects and a reset button for animation control.
Technical Description
Architecture:
Spark splat rendering with dynamic object modifiers
GLSL shader effects via Dyno system
Three.js scene with interactive camera controls
lil-gui interface for real-time parameter control
Effects:
Twister: Rotational vortex that deforms geometry with smoothstep-based animation
Rain: Falling droplet simulation with vertical displacement and scale modulation
Key Features:
Time-based animation with automatic reset on effect change
Splat loading synchronization to prevent timing issues
Real-time shader parameter updates


dmarcos:
Thanks so much. You rock!

PR #150, opened 2025-08-04
Improve .ply parsing speed using compiled parser function

mrxz:
Parsing of `.ply` files is bottlenecked by the dynamic nature of each element type. The order of the properties for a given element are determined by the order in which they appear in the header. Since this can vary from file to file the parsing has to account for this.

Currently this is done by constructing an array of parser functions (`parsers`) that handles each property in order. Despite this array being constant while parsing the items, it can't really be optimized by the browser. This PR adds a method that constructs the _source_ for a dedicated parse function, allowing the browser to compile and optimize it. This trick is based on a similar approach used in PDF.js.

Additionally a large portion of time is spent on parsing spherical harmonics. This is mainly due to the key-based lookups, which add up considerably the more splats the file contains. Treating the `f_rest_[0-9]{1,2}` properties differently and placing them directly in an array, further improves performance.

Quick comparison for a 1M splat file (compressed/SuperSplat .ply format):
| State | Time |
|--|--|
|Before | 3.54 s |
|Compiled parse function | 1.71 s |
|`f_rest` array | 0.84s |

Since the `.ply` files should be considered user-input, it's important that the constructed source code is safe. Constraints are placed on the property names and types (as these are interpolated into the generated source). In case any of these constraints are broken, the old approach is used.

The usage of `new Function` does mean that users that want to use CSP will need to allow `unsafe-eval`. However this is already the case due to the inlined worker source code.

### DIFF
--- a/src/ply.ts
+++ b/src/ply.ts
@@ -203,11 +203,7 @@ export class PlyReader {
     for (const elementName in this.elements) {
       const element = this.elements[elementName];
       const { count, properties } = element;
-      const item: Record<string, number | number[]> = {};
-      // Prepare item
-      for (const [propertyName, property] of Object.entries(properties)) {
-        item[propertyName] = property.isList ? [] : 0;
-      }
+      const item = createEmptyItem(properties);
       // Construct a parse function
       const parseFn = createParseFn(properties, this.littleEndian);
 
@@ -231,9 +227,9 @@ export class PlyReader {
     const ssChunks: SSChunk[] = [];
 
     let numSh = 0;
-    let sh1Props: string[] = [];
-    let sh2Props: string[] = [];
-    let sh3Props: string[] = [];
+    let sh1Props: number[] = [];
+    let sh2Props: number[] = [];
+    let sh3Props: number[] = [];
     let sh1: Float32Array | undefined = undefined;
     let sh2: Float32Array | undefined = undefined;
     let sh3: Float32Array | undefined = undefined;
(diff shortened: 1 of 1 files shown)


dmarcos:
It looks great to me! Thanks

asundqui:
@mrxz this is a really amazing speedup! A couple of thoughts:

1) Is it really true that we already require `unsafe-eval` for inline worker source? It looks like worker inline is done by creating a Blob URL and that may be controlled by `worker-src` (or some other fallbacks)? I'm not super deep with this stuff but I worry about enabling `unsafe-eval` - I could see users not being willing/able to allow that.

2) If the above if true, I'm wondering if it's possible to add an additional dynamic check to see if `unsafe-eval` is available? Like for example what if we added this to the beginning of `safeToCompile()`:
```
  try {
    const testUnsafeEval = new Function("return 42;");
  } catch (e) {
    return false;
  }
```
I haven't tested the above with `unsafe-eval` disabled but GPT seems to think it will simply throw an exception if not allowed.

WDYT?

mrxz:
> 1. Is it really true that we already require `unsafe-eval` for inline worker source? It looks like worker inline is done by creating a Blob URL and that may be controlled by `worker-src` (or some other fallbacks)? I'm not super deep with this stuff but I worry about enabling `unsafe-eval` - I could see users not being willing/able to allow that.

You're right that the worker source is loaded through a Blob, so instead of `worker-src 'unsafe-eval'` it would be `worker-src blob:`, though that's effectively the same ([4.2.2.1. Security Considerations for GUID URL schemes](https://www.w3.org/TR/CSP2/#source-list-guid-matching)):
> Especially for the [default-src](https://www.w3.org/TR/CSP2/#default_src) and [script-src](https://www.w3.org/TR/CSP2/#script_src) directives, policy authors should be aware that allowing "data:" URLs is equivalent to unsafe-inline and allowing "blob:" or "filesystem:" URLs is equivalent to unsafe-eval.

That being said, this PR requires it on the `script-src` directive, instead of a more limited directive like `worker-src`. Currently `script-src` needs to include `wasm-unsafe-eval` for Spark but that would now become `unsafe-eval`. So this does make the required policy less strict. My point was more that when using CSP you'd generally want neither, so this change doesn't make things markedly worse, though admittedly that's an oversimplification. 

> I haven't tested the above with unsafe-eval disabled but GPT seems to think it will simply throw an exception if not allowed.

That should indeed work. Instead of testing it each time, the test can be performed once in the global scope.

Probably a good idea to document the requirements somewhere. Otherwise users setting up CSP policies might end up whitelisting things until no errors remain, in which case they've unconsciously opted out of this usage (which they might or might not have wanted)

asundqui (approved):
Fantastic, let's get this in! People will appreciate this big speed bump in parsing.

dmarcos:
thanks so much. great effort

PR #155, opened 2025-08-14
Make sure all examples resize when the window is resized

mrxz:
Most of the examples set their size once. This causes issues when resizing the window or changing orientation on a handheld device. This PR adds a `resize` event handler to all examples. The structure follows the one used by Three.js in their examples and was already present in the interactivity example ([examples/interactivity/index.html#L71-L76](https://github.com/sparkjsdev/spark/blob/0edfc8d9232b8f6eb036d27af57dc40daf94e1f3/examples/interactivity/index.html#L71-L76))

Some examples already handled resizing, these are left as-is.


dmarcos:
Thanks so much!

PR #156, opened 2025-08-14
Ensure only one update timeout is scheduled at a time and call `gl.flush()` to encourage eager execution

mrxz:
While investigating the way CPU/GPU work is scheduled in Spark I noticed that the GPU work arising from `SparkRenderer.updateInternal` was not immediately executed, but instead waited for the next rAF or flush. See the following screenshot:
image

This PR introduces a `gl.flush()` to encourage eager execution. See [WebGL_best_practices#flush_when_expecting_results](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#flush_when_expecting_results). The browser can now immediately start the GPU work. Additionally the logic was reworked to:

- Ensure there's only a single timeout scheduled. Multiple timeouts would be scheduled in case >1 rAF took place before it got a chance to run or during WebXR sessions (once for each eye).
- Use a single `THREE.Matrix4` instance to avoid cloning `matrixWorld` each time (avoids a memory allocation).

With these changes the timeline looks as follows:
image

Edit: for clarity the "GPU idle" and "GPU work" refer to the GPU thread. The key is that no GPU work is _scheduled_ in the before case, whereas it is with this PR. Whether or not the GPU is actually idle at that point obviously depends on other factors. But in case it _is_, it can immediately start on it.

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -267,10 +267,11 @@ export class SparkRenderer extends THREE.Mesh {
   viewpoint: SparkViewpoint;
 
   // Holds data needed to perform a scheduled Gsplat update.
-  private pendingUpdate: {
-    scene: THREE.Scene;
-    originToWorld: THREE.Matrix4;
-  } | null = null;
+  private pendingUpdate = {
+    scene: null as THREE.Scene | null,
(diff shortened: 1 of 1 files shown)


dmarcos:
Great find! Thanks!

Issue #84, opened 2025-07-01
Splat not visible when using an OrthographicCamera


pikilipita:
Hi,

currently splats are not visible when rendered via an orthographic camera (THREE.OrthographicCamera), do you plan to support Orthographic cameras in future releases?

dmarcos:
@pikilipita thanks for filing. what’s your use case?

pikilipita:
Hey Marcos,
I'd like to display gaussian splats in a web video editor such as [https://pikimov.com](https://pikimov.com).

As an example, the statue model could be a gaussian splat here:
<img width="756" alt="Image" src="https://github.com/user-attachments/assets/33063393-1958-4858-9274-579e2abc2000" />

dmarcos:
do you need an ortho camera there or just rendering the splat from a different point of view?

pikilipita:
In an editor, mixing objects rendered in an orthographic way with objects rendered as perspective camera would not work: this is not something you would see in Blender, Unity...

seppestaes:
Hi @dmarcos , would be a great feature for precision editing. Best regards

kfarr:
closed by #157 @dmarcos

PR #157, opened 2025-08-15
Add support for orthographic rendering

mrxz:
Related: #84 

This PR adds support for orthographic rendering. Three.js already provides a `isOrthographic` uniform that can be used to distinguish between perspective and orthographic projection matrices.
image


dmarcos:
Thank you!

Issue #107, opened 2025-07-10
Multiple canvas/instances breaks


lucasoptml:
First of all this is a fantastic library!!  Especially excited by SOGS support and depth of field/relighting effects.

Using 0.1.5, I notice that when I try to load two SplatMesh({url}} on different canvases it breaks on the 2nd+ canvases.

Is there some global context/cache that's blocking things?  I saw something about passing bytes instead of url to the SplatMesh.

I'm really just replicating the same component multiple times (using a react three fiber <primitive> to load it).

I'm using next so navigating pages also breaks it (the component is on both pages), so I can't really use it yet.

I am using the nextjs fix so I'm not sure if that's related (I would guess it is not).

dmarcos:
Duplicate of https://github.com/sparkjsdev/spark/issues/98

Can you use iframes?

lucasoptml:
Well, no, I can't control that, so I'm stuck on mkkellogg until solution =(

dmarcos:
> Well, no, I can't control that, so I'm stuck on mkkellogg until solution =(

There's a workaround [here](https://github.com/sparkjsdev/spark/issues/107#issuecomment-3059223637). Would that work?

61cygni:
Do you have a repro of the problem we can check out?

lucasoptml:
It's just using multiple canvases and each has their own SplatMesh loaded and added independently, no overlap.  I don't think the workaround was linked so not sure.

mrxz:
> It's just using multiple canvases and each has their own SplatMesh loaded and added independently, no overlap. I don't think the workaround was linked so not sure.

@lucasoptml For convenience Spark automatically sets some things up in a Scene when you only add a `SplatMesh`. However this logic doesn't account for multiple canvasses/contexts yet. That might be the cause of the issue you're facing.

To work around it you can try manually adding a `SparkRenderer` to each of your scenes:
```js
const scene = new THREE.Scene();
const renderer = new THREE.WebGLRenderer();

const sparkRenderer = new SparkRenderer({renderer});
scene.add(sparkRenderer);

// .. Load and add SplatMeshes
```

lucasoptml:
Thanks, I'll try this soon!

lucasoptml:
Thanks, I'll try this soon!

asundqui:
[...]

Issue #98, opened 2025-07-07
Splats in diferent Canvas scopes issue


cyango:
I'm trying to re use the R3F component that has the SplatRenderer. It shows on my main canvas, but in my case I also have a modal that shows other details including a different Canvas component where I'm reusing the the Splat component, but somehow the splat is added to the main canvas not the one in the modal which is a totally different canvas context. 

Hope you're understanding the issue.

dmarcos:
Can you share a link with code to reproduce? otherwise hard to help / investigate. thanks!

asundqui:
Hi @cyango I'm not 100% sure if I understand, but it sounds like you have two <canvas> elements, and only one is rendering the SplatMesh? If that's the case, unfortunately it's a limitation of WebGL2: Each <canvas> has a separate, independent WebGL2Context, and so any uploaded splat data (which are contained in textures) can only be uploaded and used on one of the contexts. The two contexts cannot communicate, so there is no way to share the GPU data between two separate <canvas> elements.

If you'd like to make a separate SplatMesh in each context, you can do that by making a copy of the `fileBytes`, for example you can do `fileBytes.slice()` which will make a copy of the encoded PLY bytes or whatever file type you have. Then each SplatMesh in a separate canvas can decode its own copy and render its own copy in its separate WebGL context.

I hope that makes sense, does that answer your question and help you work around it?

dmarcos:
Can use an iframe in your modal?

joaomelorodrigues:
> Can use an iframe in your modal?

Well no because I'm using this as React components, wouldn't be a good practice.

(Sorry for the delay)

dmarcos:
Not React but here a vanilla JS / HTML example of two Spark instances rendered in two separate canvases in the same page:

https://stackblitz.com/edit/sogs-zshfrmnu?file=index.html

Does it help?

Issue #78, opened 2025-06-29
<SparkRenderer>  is optional but removing makes the splat disappear (R3F)


joaomelorodrigues:
In the docs I'm interpreting that <SparkRenderer>  is optional, although if i remove it my splat disappears. In the r3f example it is explicitly there,

dmarcos:
cc @dylanhu7

asundqui:
@cyango the truth if `SparkRenderer` is required in the scene, because it's what actually is "rendered". However, in the docs we say it's optional because we have some hooks when you add a `SplatMesh` that checks if there's already a SparkRenderer in the scene and if not it adds one.

So, you don't have to add one explicitly, but if you do add one (presumably you need it to tweak some of its knobs) you should definitely not remove it :). Does this make sense?

dmarcos:
@cyango are you removing the renderer at runtime? if you can share code that illustrates we will be able to better assist. thanks!

joaomelorodrigues:
> [@cyango](https://github.com/cyango) are you removing the renderer at runtime? if you can share code that illustrates we will be able to better assist. thanks!

No, I'm just commenting it out and the splat doens't show, although the onLoad callback runs on SplatMesh

dmarcos:
@cyango can you share the code? will be much easier to investigate. thanks

joaomelorodrigues:
> [@cyango](https://github.com/cyango) can you share the code? will be much easier to investigate. thanks

I'm testing this here: https://github.com/cyango/spark-react-r3f

timothyallan:
This also seems to be the case with SolidJS and a vanilla JS usage as well?
```
 const splatMesh = new SplatMesh({
          editable: false,
          url: path});
scene.add(splatMesh);
```
Will not show anything, nor is there a renderer in the scene. As soon as I add 
```
 const sparkRenderer = new SparkRenderer({ renderer });
scene.add(sparkRenderer);
```
before the mesh, everything magically shows up.

dmarcos:
@timothyallan what version? are you using 0.1.8?

timothyallan:
@dmarcos  yep, `"@sparkjsdev/spark": "^0.1.8"`

Things like `onFrame` in the `SplatMesh` don't even fire without manually adding the `SparkRenderer`, so the SplatMesh is just not in the render loop at all it seems. I'm going to dig into what kind of hooks are being used to check and see if I'm doing anything wonky that would prevent detection.

mrxz:
[...]

PR #158, opened 2025-08-15
Use child mesh in `SplatMesh` to auto-inject `SparkRenderer` instead of monkey-patching

mrxz:
The current monkey-patching of `THREE.Scene` to auto-inject a `SparkRenderer` has many limitations and can easily break. The following setups cause issues that aren't immediately clear to the user:

- More than one `THREE.Scene` (https://github.com/sparkjsdev/spark/issues/78#issuecomment-3151948004)
- Using more than one canvas (#98, #107)
- Adding a `SparkRenderer` or `SplatMesh` using [`.attach()`](https://threejs.org/docs/#api/en/core/Object3D.attach) instead of [`.add()`](https://threejs.org/docs/#api/en/core/Object3D.add)
- Adding a `SplatMesh` as an indirect descendant of the scene, if one of the ancestors has already been added to the scene.
- Adding a `SparkRenderer` as an indirect descendant of the scene
- Using a custom `scene.onBeforeRenderer`

This PR introduces an alternative approach. Each `SplatMesh` adds an empty mesh to itself. During rendering it checks if the scene contains a `SparkRenderer` and if not injects one. Regardless of the outcome it will remove itself, ensuring the check is a one-time cost per `SplatMesh`.

This does mean that the following pattern is no longer allowed:
```js
// Never add it to the scene ourselves, yet expect it to take effect
```
Though, personally I think that's for the better. A few examples used this pattern, though it's unclear to me if this was intentional or not.

### DIFF
--- a/examples/editor/index.html
+++ b/examples/editor/index.html
@@ -79,6 +79,7 @@
     const canvas = document.getElementById("canvas");
     const renderer = new THREE.WebGLRenderer({ canvas });
     const spark = new SparkRenderer({ renderer });
+    scene.add(spark);
 
     function handleResize() {
       const width = canvas.clientWidth;
--- a/examples/envmap/index.html
+++ b/examples/envmap/index.html
@@ -54,6 +54,7 @@
 
     // Explicitly create a SparkRenderer to render environment maps
     const spark = new SparkRenderer({ renderer });
+    scene.add(spark);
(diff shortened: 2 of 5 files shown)


dmarcos:
Cool! Much better than my solution. Thanks!

PR #159, opened 2025-08-15
Pre-compute lookup tables when parsing SOGS files

mrxz:
SOGS files are loaded from images with 8-bit channels. This means that for many of the properties there are only 256 distinct values. Since `numSplats` >> 256, it can be beneficial to (pre)compute a lookup table and use it. This PR implements lookup tables for the scales, quaternions, RGBA and spherical harmonics properties.

Testing with the sutro.zip the parsing is a modest ≈10% faster with these changes.


dmarcos:
Thank you!

PR #160, opened 2025-08-18
Avoid allocating THREE.Quaternion instances in `setPackedSplat`

mrxz:
Follow-up for https://github.com/sparkjsdev/spark/pull/147.

The `setPackedSplat` function still instantiated a new `THREE.Quaternion` each time. The above PR did update it for `setPackedSplatQuat`, but missed this instance.


dmarcos:
Excellent thanks!

PR #161, opened 2025-08-18
Use `Float16Array` for to/fromHalf utilities if supported

mrxz:
Similar to what is done for `floatBitsToUint`/`uintBitsToFloat` the `toHalf`/`fromHalf` can be implemented using a `Float16Array` and `Uint16Array` pair. Since support for `Float16Array` is relatively new (https://caniuse.com/mdn-javascript_builtins_float16array) this PR checks if it's present, otherwise falling back to the current JS implementation.

There's a performance uplift with this approach. Though since the `toHalf`/`fromHalf` is only used for encoding and decoding the center of splats, it's overall impact on parsing time is limited.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -15,6 +15,11 @@ import { unindent } from "./dyno/base.js";
 
 const f32buffer = new Float32Array(1);
 const u32buffer = new Uint32Array(f32buffer.buffer);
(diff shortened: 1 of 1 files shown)


dmarcos:
Thank you!

PR #163, opened 2025-08-24
Restore multi-object add support for THREE.Scene.add (#162)

### LINKED ISSUE #162: Can't add multiple objects using THREE.Scene.prototype.add
A custom override on THREE.Scene.prototype.add breaks the engine's native ability to add multiple objects in a single call. When scene.add(object1, object2) is used, only the first object is added to the scene, and the rest are ignored.

### Cause
The override function's signature accepts only a single argument (function(object)), so it discards any additional arguments passed to it before calling the original scene.add method.

### Proposed Fix
Update the override to use rest parameters (...objects) to capture all arguments into an array. Then, iterate over the array for any custom logic and use originalAdd.apply(this, ...objects) to pass all objects to the original method.
Or alternatively, follow three.js' object.add to call add on each argument: https://github.com/mrdoob/three.js/blob/54ac263593c81b669ca9a089491ddd9e240427d2/src/core/Object3D.js#L720

nsalminen:
Updates the Scene.add override to handle multiple objects in a single call to preserve original three.js behavior. Uses rest parameters to capture all arguments and applies checks for each object.

Fixes #162

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -60,10 +60,12 @@ function containsSplatMesh(object3D: THREE.Object3D) {
 }
 
 const sceneAdd = THREE.Scene.prototype.add;
-THREE.Scene.prototype.add = function (object: THREE.Object3D) {
-  hasSplatMesh = hasSplatMesh || containsSplatMesh(object);
-  hasSparkRenderer = hasSparkRenderer || object instanceof SparkRenderer;
-  sceneAdd.call(this, object);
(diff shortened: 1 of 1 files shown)


dli7319 (approved):

dmarcos:
Thanks for the patience. We removed `monkey-patching` in https://github.com/sparkjsdev/spark/pull/158 so this PR is no longer applicable. Is this still an issue on master? Thanks

dmarcos:
Closing for now but can reopen if necessary. Thanks

PR #165, opened 2025-09-09
Splat brush painting / brush erasing example

winnie1994:


dmarcos:
Thanks so much. Great addition

PR #168, opened 2025-09-14
Splat Transition Effects with Effect selector (spherical/explosion)

kali-shade:


dmarcos:
Thanks. Can you remove changes in `dist` directory? Also I think there was a `er.name kali-shade` file committed by mistake. Thanks so much

PR #175, opened 2025-09-23
Fix effect switching race: add async disposal guards

kali-shade:
Fixed effect switching race: added async disposal guards and scene cleanup to prevent stray meshes


dmarcos:
Thank you!

Issue #169, opened 2025-09-14
SOG format not working in Viewer


cyango:
Hi, i started testing this new SOG format, which i can convert from .ply to .sog using playcanvas splat-transform tool.

Then i dropped the .sog file into the viewer and it gives an error.

I saw some PRs here that added support to SOG, am I right? If so is there a way to make it work in the viewer already?

Best regards

vinneyto:
Hello. I tried recently - it works with limitations

you can try splat-transform@0.6.0

.ply to meta.json and zip (it's so called SOG v1 but latests splat-transform returns SOG v2)

https://github.com/playcanvas/splat-transform/releases/tag/v0.7.0

joaomelorodrigues:
Using splat-transform v0.8.1 it doesn't work at all for me.

<img width="404" height="206" alt="Image" src="https://github.com/user-attachments/assets/627d1def-218a-4774-8154-ab1d8d4da487" />

 So it means the viewer doesn't support the SOG v2?

dmarcos:
There's not yet support for SOG v2. Is there an option to still sport SOG v1? Also PRs welcome if someone wants to support SOGS v2. Thanks

joaomelorodrigues:
Is this the SOG v2 we're talking about?

https://blog.playcanvas.com/playcanvas-open-sources-sog-format-for-gaussian-splatting/

dmarcos:
Yes. Also see https://github.com/playcanvas/splat-transform/issues/38

dmarcos:
Mmmh not sure how much that latest announcement overlaps with the previous SOGS v2. A bit confusing. In any case any PR to support additional formats are super welcome

Alecapra96:
@cyango can use https://colab.research.google.com/drive/1lYHsfMQR97cjjXUUPL3GRrbQ7CjdgSjy?usp=sharing

nilsingwersen:
Since PR #179 got merged, are you creating a new release soon?

dmarcos:
@nilsingwersen no ETA for new version but can use main branch builds now. see CDN links in commit messages: https://github.com/sparkjsdev/spark/commits/main/

dmarcos:
The viewer should work now. anyone can confirm?

gaoxingfa:
> The viewer should work now. anyone can confirm?

confirmed

joaomelorodrigues:
Confirmed.

PR #179, opened 2025-09-25
Add support for loading SOGSv2/SOG

### LINKED ISSUE #177: Support SOG v2 format
It would be great for gsplat.js to add support for SOG v2 (or v1) 3dgs scenes since it results in higher compression ratios than niantic spz. 

For reference,
 - here is the [SOG v2 spec](https://developer.playcanvas.com/user-manual/gaussian-splatting/formats/sog/) by Playcanvas and associated [release post](https://blog.playcanvas.com/playcanvas-open-sources-sog-format-for-gaussian-splatting), 
 - and here is the [w-m/ffsplat](https://github.com/w-m/ffsplat/tree/main/src/ffsplat/conf/format) implementation of decoding/encoding original sog-v1 by Wieland Morgenstern

Edit: similar issue requested in gsplatjs here https://github.com/huggingface/gsplat.js/issues/120

mrxz:
Resolves #177 and #169 

This PR implements support for SOG files (either bundled `.sog` files or as separate files). The `pcsogs.ts` loader is extended to detect V2 files and act accordingly. As a happy coincidence the lookup tables implemented in #159 made implementing the codebooks used in SOGS v2 straightforward.

### DIFF
--- a/examples/editor/index.html
+++ b/examples/editor/index.html
@@ -753,7 +753,8 @@
         file.name.toLowerCase().endsWith('.spz') ||
         file.name.toLowerCase().endsWith('.splat') ||
         file.name.toLowerCase().endsWith('.ksplat') ||
-        file.name.toLowerCase().endsWith('.zip')
+        file.name.toLowerCase().endsWith('.zip') ||
+        file.name.toLowerCase().endsWith('.sog')
       );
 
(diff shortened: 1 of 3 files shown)


dmarcos:
Thank you!

PR #181, opened 2025-09-25
Use DecompressionStream for decompressing .spz files

mrxz:
This PR replaces the usage of `fflate` with [`DecompressionStream`](https://developer.mozilla.org/en-US/docs/Web/API/DecompressionStream) for decompressing `.spz` files. This lets the browser handle the decompression, yielding a noticeable improvement in loading times (measured in Chromium):

| Example | Main | DecompressionStream |  |
|--------------|---------|-----------------------------------|---------|
| Hello World | 341 ms | 177 ms | -48% |
| WebXR | 404 ms | 206 ms | -49% |
| Splat Transition | 284 ms | 127 ms | -55% |
|                              | 378 ms | 195 ms | -48% |
|                              | 837 ms | 470 ms | -44% |

:warning: The Compression Streams API is inherently asynchronous, but `Gunzip` from `fflate` allowed for synchronous usage. This means that the interface for the `SpzReader` had to change. Since it's part of the [exported symbols](https://github.com/sparkjsdev/spark/blob/6f327a716ff89c47c53901faac14161b3a6af7fb/src/index.ts#L16), this impacts all user-code that use `SpzReader` directly.

**Before**
```js
const spz = new SpzReader({ fileBytes });
const numSplats = spz.numSplats;
// ... (allocate resources)
spz.parseSplats( ... )
```

**After**
```js
const spz = new SpzReader({ fileBytes });
await spz.parseHeader(); // Required to access properties like `numSplats`
const numSplats = spz.numSplats;
// ... (allocate resources)
await spz.parseSplats( ... )
```

### DIFF
--- a/src/spz.ts
+++ b/src/spz.ts
@@ -18,21 +18,28 @@ export class SpzReader {
   fileBytes: Uint8Array;
   reader: GunzipReader;
 
-  version: number;
-  numSplats: number;
-  shDegree: number;
-  fractionalBits: number;
-  flags: number;
-  flagAntiAlias: boolean;
-  reserved: number;
-  parsed: boolean;
+  version = -1;
+  numSplats = 0;
+  shDegree = 0;
+  fractionalBits = 0;
+  flags = 0;
+  flagAntiAlias = false;
+  reserved = 0;
+  headerParsed = false;
+  parsed = false;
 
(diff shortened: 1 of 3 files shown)


dmarcos:
thanks

PR #183, opened 2025-09-29
Fix incorrect return type for tryPcSogsZip

mrxz:
A small mistake slipped in with #179. The return type of `tryPcSogsZip` wasn't updated to include the `PcSogsV2Json` possibility. This PR fixes that


dmarcos:
Thank you!

PR #184, opened 2025-09-29
Add `minPixelRadius` to skip splats below a given pixel radius

mrxz:
This PR introduces a `minPixelRadius` property similar to the existing `maxPixelRadius`, but instead of clamping it's used for discarding. This allows very small splats to be discarded during the vertex shader.

Worth noting that the pixel radius is checking against the radius _after_ blur has been added. This does mean that when blurring splats have a minimum size and changing the blur amount impacts this minimum size. As a result the `minPixelRadius` isn't entirely independent of the blur amount. (The same of course applies to `maxPixelRadius`, but is less relevant there)


dmarcos:
Thank you!

PR #194, opened 2025-10-14
Add interactive-ripples example

kali-shade:
Demonstrates interactive ripples effect on 3D Gaussian splats. Click anywhere on the valley scene to spawn animated ripples that distort geometry and colors. Uses Dyno shaders for real-time effects and includes camera controls (WASD + mouse)


dmarcos:
Beautiful. Thanks so much!

PR #197, opened 2025-10-21
Fix `build_rust_wasm.sh` script (correct shebang, don't require rustup)

Philipp-M:
`#!/bin/bash` is non-standard, `#!/usr/bin/env bash` should be safer as it's standardized, as `/bin/bash` for instance doesn't exist on my system.

Additionally there are users (like me... this likely concerns a few NixOS users) who aren't using rustup, I've extended the script, such that it checks whether the rust target directory exists, in which case it's *very* likely that the required rust toolchain/target is on the system.
Similar for `wasm-pack` which might already be installed on the system without using `cargo install`.

### DIFF
--- a/rust/build_rust_wasm.sh
+++ b/rust/build_rust_wasm.sh
@@ -1,20 +1,27 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
-# Check if "rustup" tool is installed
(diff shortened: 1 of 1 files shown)


dmarcos:
Thank you!

PR #199, opened 2025-10-23
Support logarithmic depth buffer

### LINKED ISSUE #198: `logarithmicDepthBuffer` is not supported
When using a `WebGLRenderer` like this:

```js
const renderer = new WebGLRenderer({
  logarithmicDepthBuffer: true,
});
```

The depth value in the depth-buffer seems to be wrong.

I'm currently trying to find the relevant piece of code where I can correct this. But maybe someone else is faster...

Here's what I have e.g. in a different shader of mine to support this (I *think* it was copied from somewhere in three.js):

```glsl
#if defined( USE_LOGDEPTHBUF )
  // Doing a strict comparison with == 1.0 can cause noise artifacts
  // on some platforms. See issue #17623.
  gl_FragDepth = vIsPerspective == 0.0 ? gl_FragCoord.z : log2( vFragDepth ) * logDepthBufFC * 0.5;
#endif
```

Philipp-M:
Fixes #198

### DIFF
--- a/src/shaders/splatFragment.glsl
+++ b/src/shaders/splatFragment.glsl
@@ -3,11 +3,8 @@ precision highp float;
 precision highp int;
 
 #include <splatDefines>
-#include <logdepthbuf_pars_vertex>
+#include <logdepthbuf_pars_fragment>
 
-#ifdef USE_LOGDEPTHBUF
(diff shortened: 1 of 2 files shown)


dmarcos:
Thank you!

Issue #193, opened 2025-10-13
Flickering when removing splats


manuel-kroeter:
1. Trigger an origin update through rotation (required in the example but actually not necessary in my bigger application)
2. Remove splat meshes from the scene
3. You can notice strange flickering effects

[index.html](https://github.com/user-attachments/files/22887945/index.html)

https://github.com/user-attachments/assets/d861fe25-a1f8-4d76-af30-0c47f21de0a2

Note: Problem does not appear with stochastic splat rendering.

manuel-kroeter:
This is another example where you can see it more. Here, I don't even change the origin of the SparkRenderer, so it does not seem to be connected. The scene is loaded and unloaded in tiles, depending on what is in the view frustum. The flickering does not happen if I disable the unloading.

https://github.com/user-attachments/assets/07c3f5cb-449b-4af4-830a-ebb60da7ac12

dmarcos:
Fixed by https://github.com/sparkjsdev/spark/pull/200

Issue #192, opened 2025-10-13
Incorrect transformations after origin update for a few frames


manuel-kroeter:
Whenever an origin update happens, you can notice some incorrect jumps for a few frames. You can reproduce it with the attached example by pressing move camera twice.

[index.html](https://github.com/user-attachments/files/22887909/index.html)

https://github.com/user-attachments/assets/db289bab-d2de-470c-90bd-4754a95d2259

Note: Problem does not appear with stochastic splat rendering.

mrxz:
I did some debugging and it seems that when an new accumulator is allocated there might still be in-flight sorting operation associated with the previous accumulator. When the sorting completes it swaps out the `PackedSplat` texture used for rendering, causing a mismatch.

This would explain why it doesn't happen with stochastic splat rendering as there won't be any ongoing sorting operation.

manuel-kroeter:
Thanks, looking forward to a fix for this and the other issue since they are basically blocking any large-scale tiling / LOD implementation based on spark.

dmarcos:
Fixed by https://github.com/sparkjsdev/spark/pull/200

PR #200, opened 2025-10-24
Fix reference counting for SplatAccumulator during updates

mrxz:
While looking into #192 and #193 I noticed the reference counting of `SplatAccumulators` wasn't always correct. There were several ways it could go wrong, resulting in a range of issues. Luckily the resulting changeset is small, but tracking them down wasn't so easy. With these changes the ref counting should work correctly. There's still some possibilities for viewpoints to keep hold of accumulators starving the `SparkRenderer`.

The following changes were made:
 * Initialize the first accumulator with a `refCount` of 1, as the `SparkRenderer` immediately uses it as its `active` accumulator
 * Determine `originChanged` between current `viewToWorld` and the `viewToWorld` of the newly retrieved accumulator. Previously this compared to the _active_ accumulator, but the one retrieved from the pool might be older
 * Let `SparkViewpoint` keep track of its two references (`pending.accumulator` and `display.accumulator`). Before it kept a maximum of 1 reference, but since sorting happens asynchronous there could be in-flight sorting operations using an `accumulator` that was already released.
 * When the new accumulator has correspondence with the previous one, also update the `viewToWorld` and trigger a `prepareViewpoint`. Otherwise there'd be a window for a `SparkViewpoint` to have mismatched `accumulator` and `viewToWorld`

### DIFF
--- a/src/PackedSplats.ts
+++ b/src/PackedSplats.ts
@@ -464,7 +464,7 @@ export class PackedSplats {
     let maxSplats = 0;
     const mapping = splatCounts.map((numSplats) => {
       const base = maxSplats;
-      // Generation happens in horizonal row chunks, so round up to full width
+      // Generation happens in horizontal row chunks, so round up to full width
       const rounded = Math.ceil(numSplats / SPLAT_TEX_WIDTH) * SPLAT_TEX_WIDTH;
       maxSplats += rounded;
       return { base, count: numSplats };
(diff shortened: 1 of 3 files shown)


mrxz:
@manuel-kroeter FYI, this should fix both #192 and #193. Thanks for the reproduction and details, made tracking it down easier. Let me know if it resolves the issues for you.

dmarcos:
Thank you!

manuel-kroeter:
Awesome, thank you! I also spent some time the last days to find the issue but was a bit hard without knowing all the internals. I did some quick testing and it looks good so far 👍

Issue #205, opened 2025-10-27
Splats fail to update on time when SplatMesh is transformed


karlleungkm:
**Expected Behavior:**
When SplatMesh has its position modified (y += 1, for example), and a camera pointing at the SplatMesh is modified accordingly (position.y += 1, then calling lookAt on the new mesh position), the rendered footage should have the SplatMesh appear unmoving while the scene around it moves.

**Current Behavior:**
The rendered splats appear to stutter behind the camera movement, staying in its previous position for a split second before updating to the proper position. In rare cases, the splats fail to move at all with no discerning causes.
Attempting to await SparkViewpoint.prepare() in the render loop results in no noticeable differences; additionally, if the argument provided to prepare() does not include `update: false`, the function very quickly stalls with no error thrown.

**Environment:**
three 0.180.0
@sparkjsdev/spark 0.1.10

**Code Example:**
```
moveSplatAndCamera(delta: number) {
    if (!this.object || !(this.object instanceof Spark.SplatMesh)) return;
    this.object.position.y += delta;
    this.camera.position.y += delta;
    this.camera.lookAt(this.object.position);
}

preparingRender: boolean = false;
async render() {
    if (this.preparingRender) return;
    this.preparingRender = true;
    try {
        await this.sparkRenderer.defaultView.prepare({
            scene: this.scene,
            camera: this.camera,
        });
        this.renderer.clear();
        this.renderer.render(this.scene, this.camera);
    } catch (error) {
        console.error(error);
    } finally {
        this.preparingRender = false;
    }
}

animate() {
    requestAnimationFrame(this.animate.bind(this));
    this.render();
}
```

mrxz:
This is an inherent limitation of the way splats are updated in Spark. Whenever a render takes places the splats are updated asynchronously. As a result the updated state will appear one or more frames later. The camera transform on the other hand takes effect immediately, hence the `SplatMesh` lagging behind.

There is a `preUpdate` options on the `SparkRenderer` ([docs](https://sparkjs.dev/docs/spark-renderer/)) to force the splat updates to happen each render synchronously. However, updates will still show at least a frame late as the viewpoint is still updated asynchronously to allow sorting to take place. So this won't help in your case.

> Attempting to await SparkViewpoint.prepare() in the render loop results in no noticeable differences; additionally, if the argument provided to prepare() does not include update: false, the function very quickly stalls with no error thrown.

[...]

PR #207, opened 2025-10-28
Fix ref counting and originToWorld in `SparkViewpoint.prepare`

mrxz:
See #205 

The `prepare` method of `SparkViewpoint` increased the `refCount` of the active accumulator in case it wasn't used in the viewpoint's display state. However the display reference is already covered, resulting in double counting this reference. At the same time there's no guarantee the viewpoint holds _any_ reference when `sortUpdate` is called. To resolve this a temporary reference is held.

Additionally there was a bug where the `prepare` method would default to not providing an `originToWorld` matrix. This causes the `SparkRenderer` to use the current `toWorld` of the active accumulator. This works as long as the `SparkRenderer` hasn't been moved, but renders incorrectly in case it did. See following scenario:
```js
// Move camera with spark renderer
camera.position.x += 10.0;
await viewpoint.prepare({ scene, camera }); // Prepares for an outdated SparkRenderer position
renderer.render(scene, camera); // Triggers a new update, renders incorrectly
```

One thing that remains tricky with regards to `prepare` is that it directly calls `sortUpdate`, ignoring any ongoing sort operation or pending sort operation. This makes it possible for `sortUpdate` to throw an error, breaking the ref counting.

### DIFF
--- a/src/SparkViewpoint.ts
+++ b/src/SparkViewpoint.ts
@@ -302,7 +302,9 @@ export class SparkViewpoint {
     while (update ?? true) {
       // Force an update, possibly with origin centered at this camera
       // to yield the best quality output.
-      const originToWorld = forceOrigin ? this.viewToWorld : undefined;
+      const originToWorld = forceOrigin
+        ? this.viewToWorld
+        : this.spark.matrixWorld;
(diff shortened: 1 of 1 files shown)


dmarcos:
Thank you!

PR #210, opened 2025-10-30
Add an option to use linearToOutputTexel.

dli7319:
When a render target is set, threejs automatically change their shaders to output in Linear color space using linearToOutputTexel: https://github.com/mrdoob/three.js/blob/e04b9f7bd7f5b17103339d343168bfab2d6e0ace/src/renderers/WebGLRenderer.js#L2198

However, with Sparkjs, we need to manually set `encodeLinear` or else we get a washed out image like this:
image

With this option, we now have the option of setting `sparkRenderer.defaultView.useLinearToOutputTexel = true;` and the Spark fragment shader will automatically output in the correct color space the same way that three.js's built-in shaders do.

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -399,6 +399,8 @@ export class SparkRenderer extends THREE.Mesh {
       time: { value: 0 },
       // Delta time in seconds since last frame
       deltaTime: { value: 0 },
+      // Whether to use three.js's built-in linearToOutputTexel function instead of encodeLinear.
+      useLinearToOutputTexel: { value: false },
       // Whether to encode Gsplat with linear RGB (for environment mapping)
       encodeLinear: { value: false },
       // Debug flag that alternates each frame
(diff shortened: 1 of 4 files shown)


dli7319:
See https://github.com/dli7319/spark/blob/linearToOutputTexelTesting/examples/hello-world/index.html for testing

mrxz:
Changes seem correct, but I'm not too fond of introducing a new property. And there's something wasteful about converting the colour from sRGB to linear _just_ to convert it back.

How about automatically setting the `encodeLinear` uniform in `SparkRenderer` based on the `currentRenderTarget`? That's what I did in #196, see [Splat.ts#L311-L319](https://github.com/sparkjsdev/spark/blob/df1f9986f3581d0fcdde2bf0098cb21226f736ec/src/Splat.ts#L311-L319).

dli7319:
I see your PR has it covered. Thanks.

Issue #209, opened 2025-10-30
Twister and Rain effects has burrs.


LSQzzx:
<img width="1668" height="875" alt="Image" src="https://github.com/user-attachments/assets/ebcc6390-3138-4996-98e1-1c1985765e9e" />

LSQzzx:
This pr can solve this issue 
#211

dmarcos:
Fixed by https://github.com/sparkjsdev/spark/pull/211

PR #211, opened 2025-10-31
fix twister & rain effects

LSQzzx:
Also apply a spin (self-rotation) so each splat rotates about its own center.


LSQzzx:
solve this problem #209

dmarcos:
can you remove dist directory from
the PR? thanks

dmarcos:
@kali-shade thoughts?

LSQzzx:
> can you remove dist directory from the PR? thanks

removed

kali-shade:
Looks good to me.

dmarcos:
Thanks so much!

PR #217, opened 2025-11-11
 [painter example] switch to coloring book painting and add i/o 

winnie1994:
* Add save/load to this example
* Make coloring mode preserve original splat luminance (i.e. "coloring book" mode)

https://github.com/user-attachments/assets/bb3a1dc3-1edc-438e-bd3c-3246b0d06422


dmarcos:
Lovely. Thanks!

PR #231, opened 2025-12-01
fix: resolve webpack/Next.js WASM data URL compatibility issue (#95)

### LINKED ISSUE #95: Nextjs import / webpack

PendingReality:
Add Vite plugin to remove unnecessary `import.meta.url` argument from data: URLs during build.

## Problem
wasm-pack generates: `new URL("data:...", import.meta.url)`

The `import.meta.url` argument is unnecessary for data: URLs and causes:
- Webpack 5: "Invalid generator object" error
- Vite: Incorrectly rewrites URL as filesystem path

## Solution
Transform during build:
- Before: `new URL("data:...", import.meta.url)`
- After: `new URL("data:...")`

This allows Next.js/webpack users to import spark normally without errors.

Fixes #95

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,8 +1,35 @@
 import fs from "node:fs";
 import path from "node:path";
+import type { Plugin } from "vite";
 import { defineConfig } from "vite";
 import dts from "vite-plugin-dts";
 import glsl from "vite-plugin-glsl";
+
+/**
+ * Vite plugin to fix WASM data URL compatibility with webpack/Next.js.
+ *
+ * wasm-pack generates code like: new URL("data:...", import.meta.url)
+ * The import.meta.url argument is unnecessary for data: URLs and causes
+ * webpack/Vite to incorrectly try to rewrite the URL as a file path.
+ *
+ * This plugin transforms:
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks so much for this!

mrxz:
This breaks the source-map. Should use [magic-string](https://github.com/rich-harris/magic-string) to make the substitution and generate the right mapping.

morsmodr:
when will these changes be released to 0.1.x package?

PR #234, opened 2025-12-06
refactor: use magic-string for proper source maps in WASM fix

PendingReality:
Improve the WASM data URL fix to use magic-string for accurate source map generation. This eliminates build warnings and ensures proper debugging support.

- Replace string.replace() with MagicString.overwrite()
- Return both code and source map from renderChunk
- Uses existing magic-string dependency (via vite-plugin-dts)

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -34,7 +34,8 @@ function fixWasmDataUrl(): Plugin {
 
       const s = new MagicString(code);
       for (const match of matches) {
-        const start = match.index!;
+        if (match.index === undefined) continue;
+        const start = match.index;
         const end = start + match[0].length;
         const replacement = `new URL(${match[1]})`;
         s.overwrite(start, end, replacement);


dmarcos:
What warnings are you referring to? I don't see any on my side when running `npm run build`

mrxz:
> What warnings are you referring to? I don't see any on my side when running `npm run build

With #231 merged the generated source map is no longer correct. Rollup detects that the plugin doesn't provide any sourcemap and logs the following warning when performing a build (twice):
```
[plugin fix-wasm-data-url] Sourcemap is likely to be incorrect: a plugin (fix-wasm-data-url) was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help
```

Do these warnings not show up for you when running `npm run build` on master?

dmarcos:
Yeah I see the warning now. Overlooked. Thanks!

Issue #237, opened 2025-12-11
How to solve the problem where memory is still occupied after the mesh is destroyed?


jianghua1016:
const packedSplats = new PackedSplats({ url });
	const splatMesh = new SplatMesh({
		packedSplats: packedSplats
	});
        const packedSplats1 = new PackedSplats({ url1 });
	const splatMesh1 = new SplatMesh({
		packedSplats: packedSplats1
	});

I loaded multiple SplatMeshes using the code above. Now, I need to remove a certain SplatMesh. Normally, I can call packedSplats.dispose(); splatMesh.dispose();
But the current problem is that the memory usage does not decrease. It seems that the destruction operation should also be performed synchronously in the SparkRenderer. However, I tried to destroy several areas but the problem didn't get solved. The key point is that I only want to remove a certain SplatMesh rather than all of them. Could you please help me figure out how to destroy it in this situation? Thank you.

dmarcos:
@mrxz any recommendation? Thanks

mrxz:
There are several points where Spark keeps a reference preventing the memory from being reclaimed. I'll see what can be done about it. For the user, calling both `packedSplats.dispose()` and `splatMesh.dispose()` should be all that needs to be done.

PR #239, opened 2025-12-15
Dispose of spherical harmonics arrays and packedArray in PackedSplats

mrxz:
Addresses #237 by expanding `PackedSplats.dispose()` to also dispose the spherical harmonics textures stored in `extra` as well as setting all references to (large) arrays to `null`. The latter is done since the `PackedSplat` and `Texture` instances themselves are still retained in various places throughout, including dyno programs. Ideally these would be GC'ed but it's impractical hunting these down. 

There is a slight change in behaviour. Previously it was possible to call `splatMesh.dispose()` and then use it again later. Generally in Three.js calling `dispose` implies the object won't be used any more and I think the same should hold for Spark for consistency. Only the `interactivity` example made use of this behaviour, but has been fixed.

That said, it was a crude way to dispose of the textures (GPU resources) while still keeping a copy of the splat data in memory. Not sure if this was intended as a way to keep VRAM usage lower in the `interactivity` example, but if so, we should probably introduce a proper method for this purpose.

### DIFF
--- a/examples/interactivity/index.html
+++ b/examples/interactivity/index.html
@@ -176,7 +176,6 @@
           if (food) food.opacity = 1 - easeInOutSine(fadeOutTime / TRANSITION_LENGTH);
         } else {
           // Fade out finished
-          if (food) food.dispose();
           fadeOutTime = null;
           // Fade in next food
           fadeInTime = 0;


dmarcos:
Thank you!

PR #244, opened 2025-12-25
Fix/pager assignment for new meshes

61cygni:
No need to merge this, but wanted highlight a bug in NewSparkRenderer.

Problem:

When dynamically adding new paged `SplatMesh` objects to a scene after the initial render setup, the meshes would fail with:

Error: PagedSplats.pager not set
    at PagedSplats.update (SplatPager.ts:180)

This happens because the pager was only assigned to meshes inside the `if (!this.pager\` block—so it only ran once when the pager was first created. Any `SplatMesh` objects added to the scene later never received the pager reference.

Solution:

Move the pager assignment loop outside the conditional block so it runs every frame when there's a pager, ensuring any newly added meshes get the pager assigned before their first render.

You can see the fix in NewSparkRenderer.ts. Feel free to cherry pick as needed. 

Also, I added a flag to turn off lod in a renderer. This keeps the scene from flickering when using portals with streaming. Eventually we should fix this issue. I suspect there is contention with two cameras.

### DIFF
--- a/src/NewSparkRenderer.ts
+++ b/src/NewSparkRenderer.ts
@@ -155,6 +155,13 @@ export interface NewSparkRendererOptions {
    */
   enableLod?: boolean;
   /**
+   * Whether to drive LOD updates (compute lodInstances, update pager, etc.).
+   * Set to false to use LOD instances from another renderer without driving updates.
+   * Only has effect if enableLod is true.
+   * @default true (if enableLod is true)
+   */
+  enableDriveLod?: boolean;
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks. Much appreciated. You could do open PR against the LOD branch. 

https://github.com/sparkjsdev/spark/tree/splat-quick-lod

It will likely be easier to understand

61cygni:
Going to resubmit to the lod branch

PR #245, opened 2025-12-25
Fix: Assign pager to new paged meshes added after initialization

61cygni:
Fixes a bug where new meshes were not getting a page table. Also includes a flag to disable lodDriving. 

Previously, the pager was only assigned to meshes when the pager was first created (inside the 'if (!this.pager)' block). This meant that new paged SplatMesh objects added to the scene after the pager was already initialized would never get the pager assigned, causing 'PagedSplats.pager not set' errors when the renderer tried to update their LOD indices.

The fix moves the pager assignment loop outside the conditional so it runs every frame when there's a pager, ensuring new meshes get the pager assigned.

### DIFF
--- a/src/NewSparkRenderer.ts
+++ b/src/NewSparkRenderer.ts
@@ -155,6 +155,13 @@ export interface NewSparkRendererOptions {
    */
   enableLod?: boolean;
   /**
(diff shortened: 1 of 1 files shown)


asundqui (approved):
Great find, I did not test that case!

PR #246, opened 2025-12-26
Add distance measurement and rescaling feature for 3DGS

MasahiroOgawa:
## Objective
Add measuring 2 point distance function. And also add rescale, save new ply function.

## Test result
I added these functions to example as below.
image

This is the new function test movie.
[Screencast from 2025-12-25 17-05-38.webm](https://github.com/user-attachments/assets/336387b7-f712-475e-87cf-43a9b873c98c)

## Summary

This PR adds a new feature for measuring distances between two points on a Gaussian splat model and rescaling the model based on a known distance.

### New Features
- **Distance Measurement**: Click two points on a 3DGS model to measure the distance between them
- **Model Rescaling**: Enter a known real-world distance and rescale the entire model
- **PLY Export**: Export the rescaled model as a standard PLY file
- **Interactive Point Adjustment**: Drag points along ray lines to fine-tune positions

### New Files
- `src/PlyWriter.ts` - PLY export library for PackedSplats data
- `examples/distance-rescale/` - Interactive example demonstrating the feature
- `test/PlyWriter.test.ts` - Unit tests for PlyWriter (15 test cases)

### Key Implementation Details
- Ray-based point selection with nearest Gaussian intersection
- Constant screen-size markers (always visible regardless of zoom)
- Billboard ring outlines for marker visibility against any background
- Proper Three.js resource disposal to prevent memory leaks

### Tests Added
- 15 unit tests for PlyWriter covering:
  - Constructor with default/custom options
  - PLY header generation validation  
  - Binary data size verification
  - Position, scale, quaternion encoding
  - Log scale encoding with zero fallback
  - Sigmoid opacity encoding with edge case clamping
  - Color DC coefficient encoding
  - Empty splats handling
  - Big endian format support
  - Multiple splats ordering

### Bug Fixes
- Fixed format string in PlyWriter: `replace()` → `replaceAll()` for proper "binary little endian" output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/distance-rescale/index.html
+++ b/examples/distance-rescale/index.html
@@ -1,0 +1,109 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <title>Spark - Distance Measurement & Rescale</title>
+  <style>
+    body {
+      margin: 0;
+      overflow: hidden;
+      font-family: Arial, sans-serif;
+      background: #000;
+    }
+
+    header {
+      position: absolute;
+      color: silver;
+      font-family: sans-serif;
+      padding: 12px 16px;
+      text-align: left;
+      width: calc(100vw - 32px);
+      pointer-events: none;
+      text-shadow:
+        2px 2px 4px rgba(0, 0, 0, 0.8),
+        -1px -1px 2px rgba(0, 0, 0, 0.6),
+        1px -1px 2px rgba(0, 0, 0, 0.6),
+        -1px 1px 2px rgba(0, 0, 0, 0.6);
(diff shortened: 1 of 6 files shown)


dmarcos:
Thanks. I'm not sure if this should be a core spark feature or leave it as a 3rd party library / component. It seems too specific

MasahiroOgawa:
Thank you for your response. I think this feature will be the reason to select spark as a default Gaussian splatting viewer, because there is no viewer in the world which can measure and rescale the model. Currently only reall3D viewer has a distance measurement function, but it can only select 2  points in 2D space, cannot rescale nor save,and it is totally in Chinese. So if this is approved, spark is the only one Gaussian splatting viewer in the world, which can measure and insert a real scale into the created Gaussian spot model.

PR #274, opened 2026-02-20
Rebuild dist with latest v2.0.0-preview

kfarr:
## Summary
- Rebuild dist files (JS + WASM) to match the current source on v2.0.0-preview
- The current dist is broken — worker.ts passes 11 args to `new_traverse_lod_trees` but the WASM only accepts 10, causing a runtime crash (`Cannot read properties of undefined (reading 'length')`)
- Also picks up: log depth buffer fix (#272), SparkRenderer rename, PLY decoder fixes, LOD chunking improvements


dmarcos:
I just rebuilt. let me know if it works

dmarcos:
@kfarr if you are importing via `package.json`. Is it working for you? in the past we saw some issues with people importing from a github branch. thanks

kfarr:
@dmarcos yes worked great, thanks!
https://github.com/3DStreet/3dstreet/pull/1481

PR #285, opened 2026-03-13
Fix: LOD camera handling causing font warnings with UI under camera (e.g. react-three/uikit Fullscreen)

joaomelorodrigues:
**Problem**

When using react-three/uikit <Fullscreen> component with <Text> and SparkSplatMesh:

driveLod used inputCamera.clone() each frame, which recursively cloned camera children (including fullscreen UI).
inputCamera.updateMatrixWorld(true) forced world-matrix updates on the entire subtree every frame.
These operations caused uikit Text components to re-resolve fonts every frame, producing repeated console warnings:

unknown font family "Gilroy Bold". Available font families are "inter". Falling back to "inter".

**Solution**

Removed camera cloning. LOD now uses a snapshot of the camera’s world matrix and position (e.g. cameraMatrixWorld.clone(), setFromMatrixPosition()).
Replaced updateMatrixWorld(true) with updateWorldMatrix(true, false) so only the camera’s matrix is updated; camera children (fullscreen UI, etc.) are no longer traversed.

**Changes**

driveLod: Snapshot camera matrix/position instead of cloning the camera.
driveLod: Use updateWorldMatrix(true, false) instead of updateMatrixWorld(true).
updateLodInstances: Accept cameraMatrixWorld and cameraPosition instead of a camera instance.
No functional change to LOD behavior; only the way camera state is passed into LOD logic.


asundqui:
Hi @joaomelorodrigues , thanks for the PR! The inputCamera cloning is unnecessary now in the driveLod code, so I removed the clone complete, simplifying it.

Here's the PR showing the removal: https://github.com/sparkjsdev/spark/commit/f9dddb4e9330fde4a58593e2096ad2f0e0dcf0a0

Could you help verify that this change fixes your issue? If so, we can close this PR.

joaomelorodrigues:
Tested on 2.0.0, and error seems to be gone. Thank you.

PR #290, opened 2026-03-31
fix: rename log-depth shader define and bump three to 0.183.2

JodokGBS:
## Summary
- Rename deprecated Three.js shader define `USE_LOGDEPTHBUF` to `USE_LOGARITHMIC_DEPTH_BUFFER` in `src/shaders/splatVertex.glsl`.
- Update `three` from `^0.178.0` to `^0.183.2` in `package.json` (`devDependencies` and `peerDependencies`).
- Refresh `package-lock.json` for the dependency bump.

## Why
`USE_LOGDEPTHBUF` was renamed upstream in Three.js. This keeps Spark compatible with newer Three.js versions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "onchange": "7.1.0",
         "spark-internal-rs": "file:rust/spark-internal-rs/pkg",
         "stats.js": "^0.17.0",
-        "three": "^0.178.0",
+        "three": "^0.183.2",
         "ts-node": "10.9.2",
         "typescript": "^5.7.3",
         "vite": "^6.0.11",
@@ -27,7 +27,7 @@
         "vite-plugin-glsl": "^1.3.1"
       },
       "peerDependencies": {
-        "three": "^0.178.0"
+        "three": "^0.183.2"
       }
     },
     "node_modules/@babel/helper-string-parser": {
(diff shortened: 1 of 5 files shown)


dmarcos:
Thanks. Is this applicable to https://github.com/sparkjsdev/spark/tree/v2.0.0-preview branch?

JodokGBS:
> Thanks. Is this applicable to https://github.com/sparkjsdev/spark/tree/v2.0.0-preview branch?

No as far as i can see, this PR is not needed for v2 as it has no #ifdef USE_LOGDEPTHBUF.
It seems in v2 Spark lets Three handle log-depth internally through its shader chunks, so there’s nothing to rename in the Spark code.

asundqui:
Hi @JodokGBS , thanks for your PR!

Indeed in v2.0.0-preview branch this is unnecessary because it doesn't use `#ifdef USE_LOGDEPTHBUF`, so there is no incompatibility there. We will be merging v2.0.0-preview into main soon so I'd prefer to only merge changes into that branch.

v2.0.0-preview has a minimum Three.js version 0.180.0. I see your PR upgrades it to a newer version, but I think we'd prefer to leave it.

There are other unrelated changes in the PR in the commit [handle nullable texture data for three 0.183 typings](https://github.com/sparkjsdev/spark/pull/290/commits/62c5ee1e0e22eaa838eb5e7ebef9408264ed497c) - I'm not sure I follow why this is needed?

JodokGBS:
Hi @asundqui 

Thanks for taking the time to look into this PR.

The [nullable texture changes](https://github.com/sparkjsdev/spark/pull/290/changes/62c5ee1e0e22eaa838eb5e7ebef9408264ed497c) were only a TypeScript compatibility fix after bumping `@types/three` to `0.183.1` ([`0b4f422`](https://github.com/sparkjsdev/spark/pull/290/changes/0b4f422a7ac08a0e4bcbcdf63d47edd076c5cd92)): strict TS started failing because `DataArrayTexture.image.data` is typed as `TypedArray | null` (so `.image.data.buffer` needs a null check). There should be no runtime behavior change.

Given that `v2.0.0-preview` is being merged into `main` soon and already doesn’t use `#ifdef USE_LOGDEPTHBUF`, I agree this PR is not needed and can be closed.

asundqui:
Closing because this is no longer needed.

Issue #287, opened 2026-03-18
Precision Issue


WilliamLiu-1997:
Hi, I am working on a project to show Gaussian Splat 3d tiles on Earth(like CesiumJS, but this project is based on ThreeJS), but because the tiles are on the surface of the earth, so it is farrrrr away from the center. And that causes the precision problem, is there any way I can do to solve this problem?
Screenshot one is using Spark, screenshot two is using renderer I write myself

<img width="2917" height="1896" alt="Image" src="https://github.com/user-attachments/assets/1d8b1010-1668-47fc-a187-6ff9ca934307" />

<img width="3838" height="1905" alt="Image" src="https://github.com/user-attachments/assets/3d88febe-12f8-4d09-b2c9-46c424a3df9d" />

dmarcos:
Can you give 2.0.0 preview branch a try?

https://sparkjs.dev/2.0.0-preview/

WilliamLiu-1997:
> Can you give 2.0.0 preview branch a try?
> 
> https://sparkjs.dev/2.0.0-preview/

I am using 2.0.0 preview already

The only solution for now is to make Camera Relative Rendering

```

const cameraWorldPos = camera.position.clone();

scene.position.sub(cameraWorldPos);
camera.position.set(0, 0, 0);
scene.updateMatrixWorld(true);
camera.updateMatrixWorld(true);

// Compute SparkRenderer's local matrix so that its matrixWorld equals
// the camera's matrixWorld.  When the parent (scene) is shifted for
// camera-relative rendering, we need to compensate for the parent's
// non-identity world matrix.
const sparkParent = this.#sparkRenderer.parent;
if (sparkParent) {
  _sparkParentWorldInverse.copy(sparkParent.matrixWorld).invert();
  this.#sparkRenderer.matrix.multiplyMatrices(
    _sparkParentWorldInverse,
    this.#camera.matrixWorld,
  );
} else {
  this.#sparkRenderer.matrix.copy(this.#camera.matrixWorld);
}
this.#sparkRenderer.updateMatrixWorld(true);

// Manually trigger SparkRenderer update so that the display
// accumulator uses the CURRENT frame's shifted coordinates.
// With autoUpdate the display lags one frame behind, which causes
// visible jitter during camera rotation because each frame's
// camera-relative shift is different.
this.#sparkRenderer.update({
  scene: this.#scene,
  camera: this.#camera,
});


render()

camera.position.copy(cameraWorldPos);
[...]

PR #294, opened 2026-04-05
improve large coordinate precision issue

WilliamLiu-1997:
Relate to https://github.com/sparkjsdev/spark/issues/287

## Summary

Improve far-from-origin precision for transformed `SplatMesh` scenes by rebasing internal splat generation, accumulation, and sorting to view-relative coordinates.

Internally, splat centers are no longer propagated as long-lived absolute world-space values by default. The pipeline now prefers centers relative to the current `viewOrigin` / accumulator origin and only converts back to world space where external semantics require it.

## What Changed

- Rebased accumulator output to `viewOrigin` for both packed and extended splat paths
- Updated depth sorting to use rebased centers
- Applied render-time origin restoration consistently in `SparkRenderer`
- Added support for generators that already emit view-relative centers
- Rebased `SplatMesh` generation before large world translations lose precision

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -728,9 +728,7 @@ export class SparkRenderer extends THREE.Mesh {
     geometry.instanceCount = spark.activeSplats;
 
     const accumToWorld = new THREE.Matrix4();
-    if (!this.display.extSplats) {
-      accumToWorld.makeTranslation(spark.display.viewOrigin);
-    }
+    accumToWorld.makeTranslation(spark.display.viewOrigin);
     const cameraToWorld = camera.matrixWorld.clone();
     const worldToCamera = cameraToWorld.invert();
     const accumToCamera = worldToCamera.multiply(accumToWorld);
--- a/src/SplatAccumulator.ts
+++ b/src/SplatAccumulator.ts
@@ -23,6 +23,7 @@ import {
   DynoUsampler2DArray,
   type DynoVal,
   DynoVec3,
+  Gsplat,
   combineCovSplat,
   combineGsplat,
   dynoBlock,
@@ -35,6 +36,7 @@ import {
   outputExtendedSplat,
   outputPackedSplat,
   outputSplatDepth,
+  select,
   splitCovSplat,
   splitGsplat,
   sub,
@@ -65,6 +67,7 @@ export class SplatAccumulator {
   static viewCenterUniform = new DynoVec3({ value: new THREE.Vector3() });
   static viewDirUniform = new DynoVec3({ value: new THREE.Vector3() });
   static sortRadialUniform = new DynoBool({ value: true });
+  static sourceRelativeUniform = new DynoBool({ value: false });
   maxSplats = 0;
   numSplats = 0;
   target: THREE.WebGLArrayRenderTarget | null = null;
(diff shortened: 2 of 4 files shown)


WilliamLiu-1997:
Here are examples：

Example that fixed precision issue
https://threejs-3d-tiles-renderer-mynn6oxso-williamliu1997s-projects.vercel.app/

Example that current Spark has (you can see model is flickering)
https://threejs-3d-tiles-renderer-7rj9n6oeu-williamliu1997s-projects.vercel.app/

WilliamLiu-1997:
Hi @asundqui, could you please take a look at this PR? If you have a better approach, I’d really appreciate your input.

This is quite critical—users need this to run Spark on the worldCoordinate project, and it’s currently blocking my progress.

Thanks a lot in advance!

asundqui:
Hi there @WilliamLiu-1997 , sorry for the slow response, we are preparing to merge v2.0.0-preview into main soon.

I think I understand what you are trying to do here, but let me try to understand:
- Have you tried `new SplatMesh({ url, extSplats: true })`, or if you're using paged splats, `SparkRenderer({ ..., pagedExtSplats: true })` ?

I want to understand if your problem is due to float16 quantization in PackedSplats, if so using ExtSplats / paged extSplats will fix that.

If that is not the issue, then the next thing you can do is try `SparkRenderer({ accumExtSplats: true })`. This will use ExtSplats for the intermediate representation used when combining multiple SplatMeshes together into the global scene.

Please can you look into those options and try them out first? If neither of these fix your problem, we can look at simplifying this PR.

WilliamLiu-1997:
> Hi there @WilliamLiu-1997 , sorry for the slow response, we are preparing to merge v2.0.0-preview into main soon.
> 
> I think I understand what you are trying to do here, but let me try to understand:
> 
> * Have you tried `new SplatMesh({ url, extSplats: true })`, or if you're using paged splats, `SparkRenderer({ ..., pagedExtSplats: true })` ?
> 
> I want to understand if your problem is due to float16 quantization in PackedSplats, if so using ExtSplats / paged extSplats will fix that.
> 
> If that is not the issue, then the next thing you can do is try `SparkRenderer({ accumExtSplats: true })`. This will use ExtSplats for the intermediate representation used when combining multiple SplatMeshes together into the global scene.
> 
> Please can you look into those options and try them out first? If neither of these fix your problem, we can look at simplifying this PR.


@asundqui  Thanks very much, I have already applied those settings, the issue still exists, by the way, the accumExtSplats=true makes it worse

The problem is that I am using spark in world coordiante system project -- model placed on the surface of the earth, which means 32 precision is still not enough, we need to calculate coordinates relative to camera position (you can see the example here [example](https://threejs-3d-tiles-renderer.vercel.app/) this example already applied this PR so there won't be precision issue i this example)

The root cause seems to be that splat centers are propagated as absolute world-space values for too long in the intermediate pipeline.

This leads to several problems:

- During object-to-world transformation, local splat centers are offset by very large world-space translations
- These large values enter float32 computations too early
- Precision is lost before reaching the final render stage
- Even if positions are later rebased relative to the camera, the lost precision cannot be recovered
- So the issue is not only in the final vertex shader, but already in earlier intermediate stages.

More specifically, the problem appears in steps such as:

- Converting splat centers from object space to world space
- Storing splat centers during accumulation
- Computing depth-sorting metrics

As long as these stages rely on large absolute world-space coordinates, precision issues will persist when the model is far from the origin.

In this PR, internally, splat centers are no longer propagated as long-lived absolute world-space values by default. The pipeline now prefers centers relative to the current viewOrigin / accumulator origin and only converts back to world space where external semantics require it.

dmarcos:
we merged v2.0.0-preview in to main. can you reopen against main? thanks!

WilliamLiu-1997:
> we merged v2.0.0-preview in to main. can you reopen against main? thanks!

Yep,done

PR #296, opened 2026-04-09
Deleted example effects that are now in effects folder

kali-shade:
Removed the effects in examples that are now in the new effects page


dmarcos:
Can you rebase?

PR #309, opened 2026-04-21
Specify 0.180.0 or higher as Three.js peer dependency

### LINKED ISSUE #308: Error resolving packages (THREE r184)
I dont know if this is actually an issue, but in my setup I have THREE r184 + Spark.js v2.0.0

Installing packages with any THREE release > 180 results in this error:
```
npm error While resolving: @sparkjsdev/spark@2.0.0
npm error Found: three@0.184.0
```

Indeed placing `^0.180.0` for THREE in package.json solves the issue, but this way THREE is downgraded to 180 while I want to use 184. Is there a specific way to solve this without using `npm i --force `?

Thanks!

mrxz:
Fixes #308 

The three peer dependency is currently constrained to `0.180.X`, whereas the intended constraint is r181 or higher. This PR changes the package.json to reflect this. While future Three.js versions might break things, introducing an upper bound is likely going to cause more issues and would require updating this every month.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "fflate": "^0.8.2"
   },
   "peerDependencies": {
-    "three": "^0.180.0"
+    "three": ">=0.180.0"
   },
   "keywords": ["3d", "three.js", "gsplats", "3dgs", "gaussian", "splats"]
 }


dmarcos:
Thanks!

PR #315, opened 2026-04-22
Save and restore activeCubeFace and activeMipmapLevel when rendering to render target

mrxz:
The `SplatAccumulator` saves and restore the current render state. However, for the render target it only recorded the target itself and not the active cube face or active mipmap level. This caused a regression in the [envMap example](https://sparkjs.dev/examples/#envmap), where the resulting envmap does not have all its faces rendered, as can be seen by the "black" reflected in the duck model.

It would be nicer if `SparkRenderer.renderCubeMap` could ensure that the splats are only updated once, instead of once per cube face. That said, this bugfix would still be needed in any situation where the current render target has a non-zero cube face or mipmap level.

### DIFF
--- a/src/SplatAccumulator.ts
+++ b/src/SplatAccumulator.ts
@@ -186,6 +186,8 @@ export class SplatAccumulator {
   private saveRenderState(renderer: THREE.WebGLRenderer) {
     return {
       target: renderer.getRenderTarget(),
+      activeCubeFace: renderer.getActiveCubeFace(),
+      activeMipmapLevel: renderer.getActiveMipmapLevel(),
       xrEnabled: renderer.xr.enabled,
       autoClear: renderer.autoClear,
     };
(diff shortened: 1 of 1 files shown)


asundqui:
We should merge this in, okay @dmarcos ?

@mrxz  I actually intended for `renderCubeMap` to not update for each cube face. I guess if `autoUpdate` is turn in you would indeed update 6 times potentially.. I usually try to steer people to have a separate SparkRenderer to use for things like rendering environment maps. In that case there is a single `.update()` that is awaited inside the renderCubeMap and the 6 renders should be from the same update step.

mrxz:
>  I actually intended for renderCubeMap to not update for each cube face. I guess if autoUpdate is turn in you would indeed update 6 times potentially.. I usually try to steer people to have a separate SparkRenderer to use for things like rendering environment maps. In that case there is a single `.update()` that is awaited inside the renderCubeMap and the 6 renders should be from the same update step.

The `envmap` example does use a separate `SparkRenderer`, but since `autoUpdate` defaults to true it still ends up with this problem. We should probably update the example to disable auto update and document this aspect of `renderCubeMap` as I can easily see users run into this issue.

Ideally I'd like to see `renderEnvMap`, `renderCubeMap` and `renderTarget` becoming obsolete as they are quite opaque to the user. It'd be nicer IMHO for users to write these how they'd normally would in vanilla Three.js with only the relevant additions needed for Spark (e.g. `.update()`).

PR #316, opened 2026-04-22
fix: fire setDirty when pager fetch pipeline drains

### LINKED ISSUE #298: onDirty chain breaks during paged LOD loading when camera is static
## Description

   When using `SparkRenderer` with paged LOD splats in an on-demand rendering architecture (render only when explicitly requested, no continuous render loop), the `onDirty` callback chain has gaps that cause splats to not fully load.

   ## Reproduction

   1. Create a `SparkRenderer` with `preUpdate: true` and an `onDirty` callback that schedules a new render frame
   2. Add a `SplatMesh` with `{ paged: new PagedSplats({ rootUrl }), lod: true }`
   3. Trigger a single render (the camera is static after this)
   4. Observe: the splat sometimes loads fully, sometimes stops partway through

   ## Workaround

   We currently poll `invalidateScene()` every 200ms while `pager.fetchers`, `pager.fetched`, `pager.newUploads`, or `pager.lodTreeUpdates` are non-empty.

   ## Potential Root Cause Analysis (AI-assisted, via Claude)

   We used Claude to trace through the compiled SparkRenderer source and it identified the following potential explanation. This is AI-assisted analysis — it may not be fully accurate, but we're sharing it in case it's helpful.

   `setDirty()` appears to be called in only 3 places inside `SparkRenderer`:

   1. **After `generate()`** in `updateInternal` — but only when `doUpdate = true`. When the camera is static and the version hasn't changed, `doUpdate = false` and this is skipped.
   2. **After sort completes** in `driveSort` — but `driveSort` returns immediately when `sortDirty = false`, which happens when `doUpdate = false` (since `sortDirty = true` is only set inside the `doUpdate` branch).
   3. **After `updateLodInstances`** in `driveLod` — but only when `tryExclusive` succeeds (LOD worker is free) AND `lodDirty = true`.

   Claude's hypothesis is that the chain breaks when all three conditions fail simultaneously:
   - Camera is static → `doUpdate = false` → no `setDirty` from (1) or (2)
   - LOD worker is busy processing a previous batch → `tryExclusive` returns null → no `setDirty` from (3)

   Meanwhile, chunks are still being downloaded by the pager's fetchers in the background. When they complete, nothing calls `setDirty()`, so no new render is scheduled to process them.

   The chain eventually resumes when the LOD worker finishes its current batch (calling `setDirty` at the end of `tryExclusive`), but the gap can be significant — especially during initial load when `initLodTree` involves network I/O for the `.rad` metadata.

   Additionally, `consumeLodTreeUpdates()` may return empty between LOD traversals even though chunks are still in flight. If `lodDirty` is false at that point, `updateLodInstances` is skipped and `setDirty` is never called.

   ## Suggested Fix (AI-assisted, via Claude)

   Consider calling `setDirty()` when paged chunk fetches complete (e.g., in `processFetched` or after `driveFetchers` starts new downloads), so that on-demand renderers are notified that new data is available for processing.

onderilkesever:
## Problem

Fixes #298.

When `SparkRenderer` is used in an on-demand rendering architecture (render only when `onDirty` fires, no continuous loop) with a static camera, paged LOD splats can stop loading partway through. The `onDirty` chain has a gap: chunk fetches complete asynchronously, outside any render frame, and nothing calls `setDirty()` to signal that new data is ready.

## Root cause

`setDirty()` is only called from:
1. After sort completes (`driveSort`)
2. When the LOD snapshot changes (`updateInternal`)
3. After `updateLodInstances` when `lodDirty = true` (`driveLod`)

All three require a render frame to be in progress. If the camera is static and `dirty = false`, `onBeforeRender` never fires → `driveLod` never runs → `consumeLodTreeUpdates` is never called → downloaded chunks sit in `lodTreeUpdates` forever.

## Fix

Add an optional `onDirty` callback to `SplatPagerOptions` and call it when `processFetched()` fully empties the `fetched` queue — i.e. after each chunk finishes decoding and is moved into the LOD pipeline. `SparkRenderer` passes `() => this.setDirty()` when constructing the pager.

This fires after **every individual chunk** lands, giving progressive rendering. The `consumeLodTreeUpdates` path is intentionally unchanged — `SparkRenderer` already calls `setDirty()` via the `lodDirty` flow when it processes those updates inside a render frame.

## Changes

- `SplatPagerOptions`: add optional `onDirty?: () => void`
- `SplatPager`: store callback, call it when `processFetched()` drains `fetched` to empty
- `SparkRenderer`: pass `() => this.setDirty()` when constructing the pager

## Verified

Confirmed against an on-demand renderer that previously required a polling workaround to drive rendering during paged LOD loading. With this fix the workaround is no longer needed.

 — Claude

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -1233,6 +1233,7 @@ export class SparkRenderer extends THREE.Mesh {
           extSplats: this.pagedExtSplats,
           maxSplats: this.maxPagedSplats,
           numFetchers: this.numLodFetchers,
+          onDirty: () => this.setDirty(),
         });
 
         const { lodId } = (await worker.call("newLodTree", {
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -1233,7 +1233,7 @@ export class SparkRenderer extends THREE.Mesh {
           extSplats: this.pagedExtSplats,
           maxSplats: this.maxPagedSplats,
           numFetchers: this.numLodFetchers,
-          onDirty: () => this.setDirty(),
+          onLodTreeUpdate: () => this.setDirty(),
         });
 
         const { lodId } = (await worker.call("newLodTree", {
--- a/src/SplatPager.ts
+++ b/src/SplatPager.ts
@@ -462,6 +462,7 @@ export interface SplatPagerOptions {
    * @default 3
    */
   numFetchers?: number;
+  onDirty?: () => void;
 }
 
 export class SplatPager {
@@ -478,6 +479,7 @@ export class SplatPager {
   autoDrive: boolean;
   numFetchers: number;
   fetchPause = 0;
+  onDirty?: () => void;
 
   splatsChunkToPage: Map<
     PagedSplats,
@@ -573,6 +575,7 @@ export class SplatPager {
 
     this.autoDrive = options.autoDrive ?? true;
     this.numFetchers = options.numFetchers ?? 3;
+    this.onDirty = options.onDirty;
 
     this.splatsChunkToPage = new Map();
     this.pageToSplatsChunk = new Array(this.maxPages);
@@ -1336,11 +1339,14 @@ export class SplatPager {
 
   private processFetched() {
     const now = performance.now();
+    let processed = false;
     while (true) {
       const fetched = this.fetched.shift();
       if (!fetched) {
+        if (processed) this.onDirty?.();
         break;
       }
+      processed = true;
       const { splats, chunk, data } = fetched;
 
       let page = this.allocatePage();
--- a/src/SplatPager.ts
+++ b/src/SplatPager.ts
@@ -462,7 +462,11 @@ export interface SplatPagerOptions {
    * @default 3
    */
   numFetchers?: number;
-  onDirty?: () => void;
+  /**
+   * Called when the LOD tree gains newly-decoded pages (the fetch queue
+   * drains), so on-demand renderers can request a redraw.
+   */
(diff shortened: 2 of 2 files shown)


mrxz:
Long overdue, but the changes look good to me overall. Though the naming could be improved. While `onDirty` is used in the `SparkRenderer` to signal when a new render is required, the `SplatPager` itself doesn't really become "dirty". A better name might be something like `onLodTreeUpdate` as that aligns with when the signal is given.

onderilkesever:
Good call — renamed the pager callback to `onLodTreeUpdate` in 00782de. Kept `SparkRenderer.onDirty` as-is since that one genuinely reflects the renderer needing a redraw. Thanks for the review!

mrxz (approved):

asundqui:
Hi @onderilkesever  (and cc: @mrxz ), thanks for your contribution. This is indeed an issue, and I've also identified some other cases that should also trigger `onDirty`. I've also been experimenting with some fuzzing techniques to ensure that everything works correctly, and also adding an `examples/on-demand` with paged RAD that people can use as a basis for doing on-demand rendering correctly.

I'm working in a separate branch that will fix this issue along with other issues I discovered, how about we hold off on this PR, and let me merge in a more comprehensive solution with testing + examples in the near future? Thanks again for your work and sorry for the delay... we should have a much more robust system within another near-term release. cc: @oscarlorentzon @dmarcos

onderilkesever:
Closing this PR as the issue will be addressed as part of the more comprehensive solution. Thanks for looking into it! @asundqui

PR #320, opened 2026-04-23
fix: recreate data textures to fix memory leak in case of scene should be disposed

### LINKED ISSUE #286: Huge memory leak in v2.0
Hi there, I found a bug with shader materials caching that leads to the accumulation of large data collections stored in program uniforms

Steps to reproduce:
1. create SplatRenderer and load first scene via SplatMesh
2. dispose SplatRenderer and SplatMesh
3. look at heap memory dump and observe big amount of data inside programMaterial var
4. repeat 1..3 to observe memory leak

minor code change `WeakMap -> Map` do this issue, as I can see

in master branch we can see
```
const programMaterial = new WeakMap<DynoProgram, THREE.RawShaderMaterial>();
```

in v2.0.0-preview branch:
```
const programMaterial = new Map<DynoProgram, THREE.RawShaderMaterial>();
```

code [here](https://github.com/sparkjsdev/spark/blob/21ab718f98189995a2e4b3fb48df70ada541f5f5/src/dyno/program.ts#L105)

Also

`SplatAccumulator.generatorProgram` Map is a static link that never cleaned up. I can clear it using direct SplatAccumulator import, but I think it can be good to see it in dispose method too.

code [here](https://github.com/sparkjsdev/spark/blob/21ab718f98189995a2e4b3fb48df70ada541f5f5/src/SplatAccumulator.ts#L357)

I can make a PR to fix theese issues as well in case of it will not brake anything

lekzd:
fixes #286 

## Stale links to old Data Textures in Dyno fix

Just recreation of DataTexture links on every dispose to prevent storing it after SparkPager.dispose() called

## Steps to check:

1. create SplatRenderer and load first scene via SplatMesh
2. dispose SplatRenderer and SplatMesh
3. look at heap memory, mind a memory value
4. repeat 1..3: memory value should not to grow

### DIFF
--- a/src/SplatPager.ts
+++ b/src/SplatPager.ts
@@ -826,34 +826,46 @@ export class SplatPager {
     this.autoDrive = false;
     this.numFetchers = 0;
 
-    this.packedTexture.value.dispose();
+    const packedTexture = this.packedTexture.value;
+    packedTexture.dispose();
+    this.packedTexture.value = SplatPager.emptyPackedTexture;
+
     if (this.extTexture.value !== SplatPager.emptyExtTexture) {
       this.extTexture.value.dispose();
     }
+    this.extTexture.value = SplatPager.emptyExtTexture;
 
     if (!this.extSplats) {
       if (this.sh1Texture.value !== SplatPager.emptySh1Texture) {
         this.sh1Texture.value.dispose();
       }
+      this.sh1Texture.value = SplatPager.emptySh1Texture;
       if (this.sh2Texture.value !== SplatPager.emptySh2Texture) {
         this.sh2Texture.value.dispose();
       }
+      this.sh2Texture.value = SplatPager.emptySh2Texture;
       if (this.sh3Texture.value !== SplatPager.emptySh3Texture) {
         this.sh3Texture.value.dispose();
       }
+      this.sh3Texture.value = SplatPager.emptySh3Texture;
+      this.sh3TextureB.value = SplatPager.emptyExtSh3BTexture;
     } else {
       if (this.sh1Texture.value !== SplatPager.emptyExtSh1Texture) {
         this.sh1Texture.value.dispose();
       }
+      this.sh1Texture.value = SplatPager.emptyExtSh1Texture;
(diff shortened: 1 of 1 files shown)


mrxz:
Thanks for the PR. Ideally we would avoid the `SplatPager` from being retained in memory, but that requires more impactful changes. In the meantime at least getting rid of the large typed arrays is already a big win.

That said, I did try the changes in the PR, but the memory was still being retained after disposing both the `SplatMesh` and `SparkRenderer`. Could you share the code you used for testing, or just the disposal logic if that works better?

lekzd:
>Could you share the code you used for testing, or just the disposal logic if that works better?

@mrxz sure! I already forget about all hacks I made on App side to fix stale links in memory:

This one is dirty patch for DynoProgram prepareMaterial method
I just copied it from original code to add `materialsMap` to clear it later

```typescript
const materialsMap = new Map<typeof spark.dyno.DynoProgram, RawShaderMaterial>();

spark.dyno.DynoProgram.prototype.prepareMaterial = function () {
  if (materialsMap.has(this)) {
    return materialsMap.get(this);
  }

  const material = new RawShaderMaterial({
    glslVersion: GLSL3,
    vertexShader: spark.utils.IDENT_VERTEX_SHADER,
    fragmentShader: this.shader,
    uniforms: this.uniforms,
  });

  materialsMap.set(this, material);

  return material;
};
```

clearance code when scene should be destroyed

```typescript
materialsMap.forEach((material, program) => {
        material.dispose();

        // Clear uniforms hard links from material cache
        Object.keys(program.uniforms).forEach((key) => {
          program.uniforms[key].value = undefined;
        });
      });
      materialsMap.clear();
```

So I tried without it and memory leak still persist

mrxz:
@lekzd Thanks for the additional details. I've created an alternative PR #326 based on this one, that nulls the texture source data instead. The general idea is still the same, allowing the large arrays to be garbage collected, but doesn't require the additional clearance code you're using.

If you could give at a try to confirm that it works, that would be appreciated.

Of course there is still some memory being leaked, though no longer huge amounts. The way the materials for `DynoPrograms` are cached should probably change such that ownership is better tracked, allowing them to be disposed properly.

Issue #321, opened 2026-04-24
Reducing bundle size and support tree-shaking


mrxz:
The bundle size has grown significantly with the 2.0.0 release. Currently weighing in at about 5MiB, compared to 578.6kB for 0.1.10. It's now also 7x the size of Three.js itself, and gzipped even 11x. While most apps using Spark will end up loading a large amount of splat data, making the bundle size less relevant, there are still benefits to reducing the bundle size. For Spark to load any splat data, it has to be downloaded, parsed and initialized, making it part of the critical request chain. In other words, savings translate to app load time as well.

Bundlephobia: [Three](https://bundlephobia.com/package/three@0.184.0) | [Spark](https://bundlephobia.com/package/@sparkjsdev/spark@2.0.0)

The goal of this issue is to discuss and collect various ways to bring the bundle size down. Supporting tree-shaking would also help for users that use a bundler. If needed, it's even possible to look into multiple entry-points for the package (e.g. all-in-one bundle and a minimal one + sub-paths for optional features).

Some possible improvements:
* [ ] #322. User can always stay on 2.0.0 or revert to 0.1.10 if needed. 
* [ ] Update `compress-to-spz.js` to use worker based loaders or deprecate the utility
* [ ] Support tree-shaking (#319):
  * [ ] Ensure the entire bundle is side-effect free (or split side-effects into a separate file)
  * [ ] Make features like `SplatSkinning`, `SplatEdits`, `SparkXr`, etc... tree-shakable
* [ ] Allow loading separate .wasm files (parallel loading, avoids Base64 inefficiencies)
* [ ] Unify both .wasm blobs into one to avoid redundancy (instead of inlining twice, use `postMessage` to pass the `Module` to the worker)
* [ ] Find ways to reduce the size of the .wasm blobs

mrxz:
The `SogsDecoder` seems to be responsible for ~37% of the resulting .wasm file size. This is mainly due to the fact that it needs both zip/deflate and webp image decoding capabilities. It might be possible to delegate these operations back to the browser, similar to how it was done in `pcsogs.ts`.

The main limiting factor is that these decoders are used in both the worker and the `build-lod` command. Only the former executes in a browser context. So two variants would have to be maintained.

mrxz:
[...]

PR #322, opened 2026-04-24
Remove `OldSparkRenderer` and related classes

mrxz:
Related to #321. The bundle effectively included two versions of Spark, the "new" and the "old" one. Since most users are expected to only ever use the new one, the old one should simply be removed. As a consequence two examples showcasing features that are no longer part of Spark have been removed as well. These were already excluded from the example page listing.

If _absolutely_ needed to keep the old files around and distributed through npm, we should really split them off into separate entry points. For example, make them accessible through something like:
```js
import {SparkRenderer} from '@sparkjsdev/spark/old';
```


dmarcos:
needs rebase

mrxz:
> needs rebase

Done, rebased this PR as well as the follow-up #335

Issue #317, opened 2026-04-22
SplatMesh.forEachSplat() returns 0 splats at onLoad when using paged .rad files


Alecapra96:
When loading a .rad file with paged: true, onLoad fires before any splat data is available. At that point:

```
splatMesh.numSplats === 0
splatMesh.packedSplats === undefined
splatMesh.isInitialized === true
splatMesh.forEachSplat() iterates 0 splats
```
This makes it impossible to compute a bounding box (or do any per-splat work) at load time for paged .rad files.

Additionally, SplatMesh.getBoundingBox() throws "Bounding box requires PackedSplats or ExtSplats" in this state, even though isInitialized is true.

<img width="1092" height="570" alt="Image" src="https://github.com/user-attachments/assets/ecdcad66-9c5d-4f66-ae17-09d8cdba34f9" />

- **Reproduction**

```
const splat = new SplatMesh({
  url: 'my-file.rad',
  paged: true,
  onLoad: () => {
    console.log(splat.numSplats);      // 0
    console.log(splat.packedSplats);   // undefined
    console.log(splat.isInitialized);  // true
    splat.forEachSplat(() => {});      // iterates 0 times
    splat.getBoundingBox();            // throws: "Bounding box requires PackedSplats or ExtSplats"
  }
});
```

- **Expected behavior**

Either:

onLoad fires only after the first chunk of splat data is available (so numSplats > 0), or
A separate callback like onFirstChunkLoaded / onSplatsReady is provided for paged files, or
getBoundingBox() and forEachSplat() work correctly once isInitialized === true, regardless of paging mode

- **Workaround**

Currently working around this by polling numSplats in onFrame:

```
const splat = new SplatMesh({
  url: 'my-file.rad',
  paged: true,
  onFrame: ({ mesh }) => {
    if (mesh.numSplats > 0 && !boundingBoxComputed) {
      // compute bounding box here
      boundingBoxComputed = true;
    }
  }
});
```

- **Environment**

@sparkjsdev/spark version: 2.0.0
File format: .rad with paged: true
THREE.js version: r180

mrxz:
Currently a `SplatMesh` that is paged is immediately initialized with 0 splats. Once the chunks are loaded and pages swapped in, its `numSplats` will increase. Regardless, neither `getBoundingBox` nor `forEachSplat` is implemented/supported for paged splats.

[...]

PR #324, opened 2026-04-24
Implement forEachSplat and getBoundingBox for PagedSplats

mrxz:
This PR moves the `forEachSplat` method into the `SplatSource` interface, ensuring it works across `PackedSplats`, `ExtSplats` and `PagedSplats`. This allows the `SplatMesh.forEachSplat` and `SplatMesh.getBoundingBox` methods to work for paged splats (see #317).

These implementations are somewhat unintuitive for paged splats, as they depend on which splat pages happen to be loaded. This might confuse users, though arguably the current behaviour of no-op for `forEachSplats` isn't intuitive either.

### DIFF
--- a/src/SplatMesh.ts
+++ b/src/SplatMesh.ts
@@ -182,6 +182,17 @@ export interface SplatSource {
   }: { index: DynoVal<"int">; viewOrigin?: DynoVal<"vec3"> }): DynoVal<
     typeof Gsplat
   >;
+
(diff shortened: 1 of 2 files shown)


asundqui:
That's great! Let's merge it I say! Surely someone will want to do this :) @dmarcos

dmarcos:
Thanks

PR #326, opened 2026-04-28
SplatPager: null texture source to ensure underlying arrays can be GC'ed

mrxz:
Alternative PR to #320 that nulls the texture source data. This removes all references to the underlying TypedArrays instead of only clearing the reference from `SplatPager -> Texture`. Since these textures are also referenced from (cached) shader materials, this avoids having to [hunt them down manually](https://github.com/sparkjsdev/spark/pull/320#issuecomment-4327227356).

Ultimately it's still a crude approach, but at least ensures no large memory leaks occur. A proper solution would be to ensure that the cached shader materials are disposed of. This way no references to the textures should remain, allowing them to GC'ed normally.

### DIFF
--- a/src/SplatPager.ts
+++ b/src/SplatPager.ts
@@ -827,32 +827,41 @@ export class SplatPager {
     this.numFetchers = 0;
 
     this.packedTexture.value.dispose();
+    this.packedTexture.value.source.data = null;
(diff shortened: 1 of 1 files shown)


lekzd:
@mrxz I confirm it fixes biggest memory leak

Issue #225, opened 2025-11-20
There is a noticeable difference in speed: Spark's orbitControl is much slower than MKK's GaussianSplats3D.


cloris:
Using the same ply of SPLAT(about 2,878,840 splats) and default threejs Control three/addons/controls/OrbitControls.js, the performance difference is obvious: Spark's orbitControl is much slower than MKK's GaussianSplats3D.

mrxz:
It's unclear to me what the relevance of the `OrbitControls` is when comparing performance. Just to be sure, you aren't talking about the movement speed of the camera/controls nor using `OrbitControls` for [rendering on demand](https://threejs.org/manual/#en/rendering-on-demand), right?

There are many factors that influence performance, so it would help if you could provide:

- Device used (including which GPU)
- Browser + version
- Spark version

And ideally provide a runnable example with the .ply in question or a substitute that exhibits the same issue.

Worth pointing out that Spark does the spherical harmonics and sorting asynchronous to the rendering. This means that large camera movements/rotations might show incorrect rendering results for a few frames. Especially on mobile GPUs (tile-based) in smartphones or Apple M series devices this sorting tends to take longer. See https://github.com/sparkjsdev/spark/issues/205#issuecomment-3463883322

Using an FPS counter should tell you whether your scene is rendering at a lower rate or not.

cloris:
@mrxz Thanks your feedback. I used Windows 11 PC (12th Gen Intel(R) Core(TM) i7-12700H  2.30 GHz 4GB GPU), Chrome 141.0.7390.123, Spark "@sparkjsdev/spark": "^0.1.10", and render continuously. Ply size is 1.288GB, splats count is 18,942,182. I recorded comparison videos — one with Spark and one with Mkk,  FPS is shown in the left-top corner. Spark performance is noticeably lower than Mkk's

Spark
https://github.com/user-attachments/assets/46a3396b-77ab-4894-aec6-39f013030330

Mkk
https://github.com/user-attachments/assets/97434200-61ed-48ba-bda5-4d1bcb1aef80

.


-

notthesame23:
First of all, thank you for developing Spark, it looks really promising feature and implementation wise and the documentation is also top notch!

 The glitches after Orbit rotation is started, can be severe depending on scene...you can see really large splats, covering approx. 30% of the screen, really giving buggy/glitchy impressions.  

[...]

PR #327, opened 2026-04-29
make sort32 fast

39ali:
try to improve the performance of sort32, on avg it's 30-40% faster .

things that changed : 

- pass 2 no longer re-reads `keys[]` , `scratch` stores a packed u64 of `(inverted_key << 32 | original_index)`. pass 2 reads the high 16 bits directly from scratch with `kv >> 48` making it a sequential scan

- histogram and scatter are now branchless to help llvm vectorize the loop 

 - manually unrolled histogram and both scatter passes to 8-wide

### DIFF
--- a/rust/spark-rs/src/sort.rs
+++ b/rust/spark-rs/src/sort.rs
@@ -102,7 +102,7 @@ impl Sort32Buffers {
 }
 
 #[inline(always)]
-unsafe fn prefix_sum_exclusive(buckets: &mut [u32]) -> u32 {
+fn prefix_sum_exclusive(buckets: &mut [u32]) -> u32 {
     let mut sum = 0u32;
     for b in buckets.iter_mut() {
         let tmp = *b;
@@ -129,6 +129,7 @@ pub fn sort32_internal(
 
     // Split buckets
     let (b0, b1) = buckets.split_at_mut(RADIX_BASE);
+    let b1 = &mut b1[..RADIX_BASE];
 
     b0.fill(0);
     b1.fill(0);
@@ -146,7 +147,7 @@ pub fn sort32_internal(
                 let r1 = inv >> RADIX_BITS;
 
                 b0[r0 as usize] += valid;
-                b1[r1 as usize] += valid;
+                unsafe { *b1.get_unchecked_mut(r1 as usize) += valid };
             }};
         }
 
@@ -164,16 +165,13 @@ pub fn sort32_internal(
         let valid = (k < DEPTH_INFINITY_F32) as u32;
         let inv = !k;
         b0[(inv & RADIX_MASK) as usize] += valid;
-        b1[(inv >> RADIX_BITS) as usize] += valid;
+        unsafe { *b1.get_unchecked_mut((inv >> RADIX_BITS) as usize) += valid };
     }
 
-    // exclusive prefix‑sum → starting offsets
-    let active = unsafe { prefix_sum_exclusive(b0) } as usize;
-    unsafe {
-        prefix_sum_exclusive(b1);
-    }
+    let active = prefix_sum_exclusive(b0) as usize;
+    prefix_sum_exclusive(b1);
 
-    // pass 1: scatter into scratch 
+    // pass 1: scatter into scratch
     let mut chunks = keys.chunks_exact(8);
     let mut i = 0;
 
@@ -184,11 +182,11 @@ pub fn sort32_internal(
                 let inv = !$k;
 
                 let r0 = (inv & RADIX_MASK) as usize;
-                let pos = b0[r0] as usize;
+                let pos = unsafe { *b0.get_unchecked(r0) } as usize;
 
                 // Always write (branchless), but only advance if valid
-                scratch[pos] = ((inv as u64) << 32) | ($idx as u64);
-                b0[r0] += valid;
+                unsafe { *scratch.get_unchecked_mut(pos) = ((inv as u64) << 32) | ($idx as u64) };
+                unsafe { *b0.get_unchecked_mut(r0) += valid };
             }};
         }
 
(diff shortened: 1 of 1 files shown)


rust/spark-worker-rs/src/sort.rs
```diff
@@ -93,84 +95,167 @@ impl Sort32Buffers {
...
+        if self.buckets.len() < RADIX_BASE * 2 {
+            self.buckets.resize(RADIX_BASE * 2, 0);
         }
     }
 }
 
-/// Two‑pass radix sort (base 2¹⁶) of 32‑bit float bit‑patterns,
-/// descending order (largest keys first). Mirrors the JS `sort32Splats`.
+#[inline(always)]
+unsafe fn prefix_sum_exclusive(buckets: &mut [u32]) -> u32 {
```
mrxz: Is there a specific reason this is marked `unsafe`? It compiles just fine without.

mrxz:
Awesome work, gave it a try and can confirm that it improves sorting performance. In my limited testing I saw ~20% reduction in sorting time (~25% faster).

> manually unrolled histogram and both scatter passes to 8-wide

Without this change the performance gain seems to be roughly the same, or at least I didn't observe any significant difference. The majority of the benefit seems to come from making it branchless.

rust/spark-worker-rs/src/sort.rs
39ali: i had many experiments with simd, which didn't make it marginally faster so i removed it for simplicity sake but forgot to remove the unsafe, will clean it up

39ali:
@mrxz  i squeezed a bit more performance ~<=1ms by removing more branches from hot loops, and what you noticed seems about right, it will differ from one wasm engine to another, and arch to another(specially cache sizes and arch) so it's hard to give a solid number but it'll still be a pump in performance

dmarcos:
can you remove the changes in the dist directory?

asundqui:
@39ali great work! This looks like a cool win indeed, thanks for the work. Could remove the build in the `dist/` folder from your branch, then we can merge it in?

mrxz:
Could the macros used for unrolling the loops also be used for the body of remainder loops? Both should be identical, so if we could avoid the duplication we avoid the risk of it ever getting out of sync.

dmarcos:
@39ali any chance for you to implement @mrxz suggestions? Thanks so much for the contribution

39ali:
I'll implement the changes

39ali:
@mrxz  @dmarcos @asundqui  done !

asundqui:
@39ali  really great work here! I've done some benchmarking on your method, and I'm actually getting 2x - 4x speedups in sorting from this. I'm truly shocked that this was possible! This will have a great impact on Spark's sorting performance. On a 10M splat scene on my M3 it goes from 250 ms to 60 ms or so. It's possible that the speedup is not as great on other environments, such as @mrxz  was reporting 25% speedups on his system.

I went through and carefully separated the optimizations and measured them:

- I found that approx 65% of the gain could be explained by storing (key, index) as a packed `u64` in `scratch`, which turned the next loop from a random gather into a sequential read. So much better cache performance as a result.

- The next 20% came from doing `unsafe { unchecked... }` array accesses where the compiler couldn't be sure that it would always be in bounds, so it has to check it every iteration. I went through the logic and it looks like it should always be in bounds.

- The final 15% or so came from unrolling the loops. I had thought the unrolling couldn't possibly help because of branch prediction + cpu instruction reordering but it all helps!

It did seem like there was one error though: the second branchless loop seems problematic... I think writing to the array and only advancing the pointer if it's "valid" could overwrite things. So I removed it. I don't think it does very much for the performance anyway.

Finally I reverted some unnecessary changes to make it closer to @mrxz 's original formulation. I think we should merge this in @dmarcos , @mrxz ! WDYT? This should really help with #225 .

Interestingly, because the sorting is so much faster, it sort of exposes the next bottleneck more: uploading the ordering frequently to the GPU can cause stuttering sometimes when the counts get large. Now this happens more often!

mrxz:
> On a 10M splat scene on my M3 it goes from 250 ms to 60 ms or so. It's possible that the speedup is not as great on other environments, such as @mrxz was reporting 25% speedups on his system.

There is bound to be some variability between setups, but that's quite the discrepancy between the measurements. My guess would be that besides the gains of being branchless and more cache friendly it somehow avoids a slow-path on the M3? Regardless, it's a net positive, even the comparatively modest speed-up I've measured is a very welcome improvement.

> It did seem like there was one error though: the second branchless loop seems problematic... I think writing to the array and only advancing the pointer if it's "valid" could overwrite things. So I removed it. I don't think it does very much for the performance anyway.

The overwriting was intentional AFAICT. Since the 'invalid' entries aren't tallied you don't want to count them when scattering either. The corollary is that by only advancing after writing a valid entry, you guarantee that the final write to the array at a given position _is_ a valid entry. The only way this goes wrong is once a bucket has been fully scattered, as then it'll point into the start of the next bucket. The inverted `Infinity` value, however, has 0xFFFF as its low bytes, always placing it in the final bucket which can safely overflow up to `max_splats`.

That said, I do think the explicit check is more readable and some quick testing doesn't show a meaningful performance difference on my end between the two.

> I think we should merge this in @dmarcos , @mrxz ! WDYT? This should really help with https://github.com/sparkjsdev/spark/issues/225 .

I don't see any reason not to merge this.

Regarding #225 it will most definitely reduce the time a stale ordering is visible on screen. Whether or not this will be enough is the question. Similar to the pesky FOUC in web-design, no matter how short it will remain an "issue". Not sure what mkkellogg did differently, but it at the very least it degenerated more gracefully. It does have logic to detect large camera changes and queues partial sorts, though even before a sort update arrives, the rendered frame somehow looks less bad (subjectively).

dmarcos:
Thanks everyone. Good stuff!

PR #331, opened 2026-05-04
Replace premultipliedAlpha property with getter/setter

mrxz:
In Spark 0.1 the `premultipliedAlpha` property could be changed on the fly. Since 2.0.0 this no longer works, as it's only propagated to the underlying material in the constructor. This PR replaces the `premultipliedAlpha` property with a getter/setter pair that uses the underlying `material.premultipliedAlpha` property directly.

Additionally some of the new properties in `SparkRendererOptions` weren't using proper JSDoc comments, these have been fixed as well.


dmarcos:
Thanks!

PR #332, opened 2026-05-04
Add SPZ v4 (NGSP / ZSTD multi-stream) read and write support

udwinj:
## Summary

Adds support for SPZ v4 to spark — both reading and writing — bringing parity with the latest [`nianticlabs/spz`](https://github.com/nianticlabs/spz) reference encoder.

SPZ v4 replaces the single gzip-wrapped payload (v1–v3) with a 32-byte `NGSP` header followed by per-attribute ZSTD-compressed streams. The wire format is identical to upstream's `saveSpz()` so files round-trip cleanly between the C++ encoder and spark.

### Dependencies
- Adds `@bokuweb/zstd-wasm` (~50 KB WASM blob, lazy-loaded). Used for both ZSTD compression and decompression. Native `CompressionStream("zstd")` was considered but dropped because it isn't yet supported in Firefox or Safari.

### `src/SplatLoader.ts`
- `getSplatFileType` recognizes the `NGSP` magic at offset 0 (v4) in addition to the existing gzip-wrapped detection (v1–v3).

### `src/spz.ts` — read path
- `SpzReader` detects v4 in the constructor by inspecting the first 4 bytes; legacy files continue to flow through `GunzipReader` unchanged.
- For v4, `parseHeader()` parses the 32-byte NGSP header, awaits ZSTD WASM init, and decompresses every attribute stream up front into `v4Streams: Uint8Array[]`.
- `parseSplats()` uses a small `read()` abstraction so v3 (gzip stream) and v4 (pre-decompressed buffers) share the same decode logic.
- The smallest-three quaternion branch now triggers on `version >= 3` (was `=== 3`) since v4 uses the same encoding as v3.
- The LOD tail (a spark-only extension to the gzip format) is correctly skipped for v4 files.

### `src/spz.ts` — write path
- `SPZ_VERSION` bumped to `4`.
- `SpzWriter` now stores each attribute in its own `Uint8Array` (positions, alphas, colors, scales, rotations, sh) and assembles `[32-byte header][TOC][ZSTD streams]` in `finalize()`.
- Setter API (`setCenter` / `setAlpha` / `setRgb` / `setScale` / `setQuat` / `setSh`) is unchanged, so `transcodeSpz` and other callers don't need updates.
- TOC entries are `[u64 compressedSize LE][u64 uncompressedSize LE]`, matching the reference encoder.

## Backward compatibility

- v2, v3 file reads are unchanged (same code path).
- v3 → v4 is the **only** writer behavior change; the writer no longer emits gzip files. Callers depending on writing legacy v3 files would need a flag added in a follow-up.

## Known gaps (out of scope)

- **Extensions** (`FlagHasExtensions = 0x2`): not read or written. Reader correctly skips over them via the `tocByteOffset` field, so files containing extensions still load.
- **SH degree 4** (upstream `SH_MAX_DEGREE = 4`): pre-existing spark limitation — `SH_DEGREE_TO_VECS` only goes up to 3. Files with `shDegree == 4` already failed to load before this PR; behavior is unchanged.

### DIFF
--- a/rust/spark-lib/src/spz.rs
+++ b/rust/spark-lib/src/spz.rs
@@ -4,6 +4,7 @@ use miniz_oxide::inflate::core::inflate_flags::{
     TINFL_FLAG_USING_NON_WRAPPING_OUTPUT_BUF,
 };
 use miniz_oxide::inflate::TINFLStatus;
+use std::io::Read;
 
 use crate::decoder::{ChunkReceiver, SetSplatEncoding, SplatGetter, SplatInit, SplatReceiver};
 use miniz_oxide::deflate::compress_to_vec;
@@ -11,19 +12,34 @@ use miniz_oxide::deflate::compress_to_vec;
 pub const SPZ_MAGIC: u32 = 0x5053474e; // "NGSP"
 const SH_C0: f32 = 0.28209479177387814;
 const MAX_SPLAT_CHUNK: usize = 65536;
+const NGSP_HEADER_SIZE: usize = 32;
+const TOC_ENTRY_SIZE: usize = 16; // [u64 compressedSize LE][u64 uncompressedSize LE]
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum SpzDecoderStage { Centers, Alphas, Rgb, Scales, Quats, Sh, Extension, ChildCounts, ChildStarts, Done }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum SpzFormat {
+    Unknown, // not yet detected (need at least 4 bytes)
+    Gzip,    // legacy v1-v3: header + payload all inside a gzip stream
+    Ngsp,    // v4: 32-byte NGSP header + TOC + ZSTD-compressed attribute streams
+}
+
 pub struct SpzDecoder<T: SplatReceiver> {
     splats: T,
+    format: SpzFormat,
+    // Gzip path state (v1-v3)
     decompressor: DecompressorOxide,
     compressed: Vec<u8>,
     decompressed: Vec<u8>,
-    buffer: Vec<u8>,
-    state: Option<SpzDecoderState>,
     gzip_header_done: bool,
     out_pos: usize,
+    // V4 path state — accumulate the entire file before processing
+    raw: Vec<u8>,
+    v4_decoded: bool,
+    // Shared: decompressed payload bytes feeding the section state machine
+    buffer: Vec<u8>,
+    state: Option<SpzDecoderState>,
     done: bool,
 }
 
@@ -31,6 +47,7 @@ impl<T: SplatReceiver> SpzDecoder<T> {
     pub fn new(splats: T) -> Self {
         Self {
             splats,
+            format: SpzFormat::Unknown,
             decompressor: DecompressorOxide::new(),
             compressed: Vec::new(),
             decompressed: vec![0u8; 128 * 1024],
@@ -39,6 +56,8 @@ impl<T: SplatReceiver> SpzDecoder<T> {
             gzip_header_done: false,
             out_pos: 0,
             done: false,
+            raw: Vec::new(),
+            v4_decoded: false,
         }
     }
 
@@ -67,8 +86,8 @@ impl<T: SplatReceiver> SpzDecoder<T> {
         }
 
         let version = read_u32_le(&self.buffer[4..8]);
-        if version < 1 || version > 3 {
-            return Err(anyhow::anyhow!("Unsupported SPZ version: {}", version));
+        if !(1..=3).contains(&version) {
+            return Err(anyhow::anyhow!("Unsupported legacy SPZ version: {}", version));
         }
 
         let num_splats = read_u32_le(&self.buffer[8..12]) as usize;
@@ -78,7 +97,19 @@ impl<T: SplatReceiver> SpzDecoder<T> {
         let _reserved = self.buffer[15];
 
         self.buffer.drain(..16);
-        let state = SpzDecoderState::new(version as u32, num_splats, sh_degree, fractional_bits, flags)?;
+        self.init_state(version, num_splats, sh_degree, fractional_bits, flags)?;
+        Ok(())
+    }
+
+    fn init_state(
+        &mut self,
+        version: u32,
+        num_splats: usize,
+        sh_degree: usize,
+        fractional_bits: u8,
+        flags: u8,
+    ) -> anyhow::Result<()> {
+        let state = SpzDecoderState::new(version, num_splats, sh_degree, fractional_bits, flags)?;
         self.state = Some(state);
 
         self.splats.init_splats(&SplatInit {
@@ -94,6 +125,91 @@ impl<T: SplatReceiver> SpzDecoder<T> {
             })?;
         }
 
+        Ok(())
+    }
+
+    /// Once we have the full v4 file in `self.raw`, parse the 32-byte NGSP header and TOC,
+    /// ZSTD-decompress every attribute stream, concatenate the decompressed bytes into
+    /// `self.buffer`, and run the existing section state machine. Idempotent — only runs once.
+    fn try_decode_v4(&mut self) -> anyhow::Result<()> {
+        if self.v4_decoded {
+            return Ok(());
+        }
+        if self.raw.len() < NGSP_HEADER_SIZE {
+            return Ok(());
+        }
+
+        let magic = read_u32_le(&self.raw[0..4]);
+        if magic != SPZ_MAGIC {
+            return Err(anyhow::anyhow!("Invalid v4 SPZ magic: 0x{:08x}", magic));
+        }
+        let version = read_u32_le(&self.raw[4..8]);
+        if version != 4 {
+            return Err(anyhow::anyhow!("Unsupported NGSP version: {}", version));
+        }
+        let num_splats = read_u32_le(&self.raw[8..12]) as usize;
+        let sh_degree = self.raw[12] as usize;
+        let fractional_bits = self.raw[13];
+        let flags = self.raw[14];
+        let num_streams = self.raw[15] as usize;
+        let toc_byte_offset = read_u32_le(&self.raw[16..20]) as usize;
+        // bytes 20..32 reserved
+
+        if toc_byte_offset < NGSP_HEADER_SIZE {
+            return Err(anyhow::anyhow!(
+                "Invalid v4 tocByteOffset: {} < {}",
+                toc_byte_offset, NGSP_HEADER_SIZE
+            ));
+        }
+        let toc_size = num_streams.checked_mul(TOC_ENTRY_SIZE)
+            .ok_or_else(|| anyhow::anyhow!("v4 TOC size overflow"))?;
+        let toc_end = toc_byte_offset.checked_add(toc_size)
+            .ok_or_else(|| anyhow::anyhow!("v4 TOC end overflow"))?;
+        if self.raw.len() < toc_end {
+            return Ok(()); // need more bytes
+        }
+
+        // Walk TOC to compute total expected file size; bail out and wait for more bytes if short.
+        let mut compressed_offsets: Vec<(usize, usize)> = Vec::with_capacity(num_streams);
+        let mut data_cursor = toc_end;
+        for i in 0..num_streams {
+            let e = toc_byte_offset + i * TOC_ENTRY_SIZE;
+            let cs_lo = read_u32_le(&self.raw[e..e + 4]) as u64;
+            let cs_hi = read_u32_le(&self.raw[e + 4..e + 8]) as u64;
+            let _us_lo = read_u32_le(&self.raw[e + 8..e + 12]) as u64;
+            let _us_hi = read_u32_le(&self.raw[e + 12..e + 16]) as u64;
+            let compressed_size = (cs_lo | (cs_hi << 32)) as usize;
+            if cs_hi != 0 || compressed_size > usize::MAX / 2 {
+                return Err(anyhow::anyhow!("v4 stream too large"));
+            }
+            compressed_offsets.push((data_cursor, compressed_size));
+            data_cursor = data_cursor
+                .checked_add(compressed_size)
+                .ok_or_else(|| anyhow::anyhow!("v4 stream offset overflow"))?;
+        }
+        if self.raw.len() < data_cursor {
+            return Ok(()); // need more bytes for the compressed streams
+        }
+        // We have everything required. Decompress every stream and concatenate into self.buffer.
+        self.buffer.clear();
+        for (offset, size) in &compressed_offsets {
+            let compressed = &self.raw[*offset..*offset + *size];
+            let mut decoder = ruzstd::StreamingDecoder::new(compressed)
+                .map_err(|e| anyhow::anyhow!("v4 ZSTD init failed: {}", e))?;
+            let pre_len = self.buffer.len();
+            decoder
+                .read_to_end(&mut self.buffer)
+                .map_err(|e| anyhow::anyhow!("v4 ZSTD decompress failed: {}", e))?;
+            let _ = pre_len; // (decompressed sizes already validated by ruzstd against frame headers)
+        }
+        self.v4_decoded = true;
+
+        // Initialize the section state machine with the parsed v4 metadata, then run it.
+        self.init_state(version, num_splats, sh_degree, fractional_bits, flags)?;
+        self.poll_sections()?;
+        // For v4 the gzip stream concept doesn't apply; mark "done" so finish() validates against
+        // section completion rather than against gzip stream-end.
+        self.done = true;
         Ok(())
     }
 
@@ -225,7 +341,7 @@ impl<T: SplatReceiver> SpzDecoder<T> {
                     }
                 }
                 SpzDecoderStage::Quats => {
-                    let bytes_per_item = if state.version == 3 { 4 } else { 3 };
+                    let bytes_per_item = if state.version >= 3 { 4 } else { 3 };
                     let avail_items = self.buffer.len() / bytes_per_item;
                     let remaining = state.num_splats - state.next_splat;
                     if (avail_items < remaining) && (avail_items < MAX_SPLAT_CHUNK) {
@@ -236,8 +352,8 @@ impl<T: SplatReceiver> SpzDecoder<T> {
                     if state.output.len() < chunk * 4 {
                         state.output.resize(chunk * 4, 0.0);
                     }
-                    if state.version == 3 {
-                        // Version 3 uses "smallest three" compression for quaternions (4 bytes per splat)
+                    if state.version >= 3 {
+                        // Version 3 and v4 use "smallest three" compression for quaternions (4 bytes per splat)
                         for i in 0..chunk {
                             let base = i * 4;
                             let comp = (self.buffer[base] as u32)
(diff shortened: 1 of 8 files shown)


udwinj:
## Update: Rust WASM decoder now also supports SPZ v4

After initial testing, we found that the viewer's load path (`SplatLoader` → `workerPool` → `worker.ts`) goes through the **Rust WASM decoder** (`rust/spark-lib/src/spz.rs`), not the TypeScript `SpzReader`. The first commit's TS changes only covered the `transcodeSpz` / legacy worker path. V4 files were failing with `"Invalid gzip header"` because the Rust decoder was hitting the NGSP magic bytes and treating them as malformed gzip.

A second commit adds full v4 support to the Rust decoder.

---

### What the second commit changes

#### New dependency — `ruzstd 0.7.3`
Pure-Rust ZSTD decoder, no C bindings, compiles cleanly to WASM.

```toml
# rust/Cargo.toml (workspace)
ruzstd = { version = "0.7.3", default-features = false, features = ["std"] }
```

#### `rust/spark-lib/src/decoder.rs`
`MultiDecoder` now recognises NGSP magic at the start of a file (in addition to the existing gzip-wrapped detection):

```rust
if magic == SPZ_MAGIC {
    // NGSP magic at file start — SPZ v4 (ZSTD multi-stream, not gzip-wrapped)
    return self.init_file_type(SplatFileType::SPZ);
}
```

#### `rust/spark-lib/src/spz.rs`

**Format detection** — new `SpzFormat` enum; the decoder self-detects on the first 4 bytes of input:

```rust
enum SpzFormat { Unknown, Gzip, Ngsp }
```

- First 4 bytes = `NGSP` → v4 path (accumulate raw bytes, decode via `try_decode_v4()`)
- First 4 bytes = gzip magic → v1–v3 path (existing streaming gzip, unchanged)

**`try_decode_v4()`** — once all bytes are buffered, parses the 32-byte NGSP header, walks the TOC, ZSTD-decompresses each attribute stream with `ruzstd`, then feeds decompressed bytes into the existing per-stage state machine (`Centers → Alphas → Rgb → Scales → Quats → Sh`):

```rust
fn try_decode_v4(&mut self) -> anyhow::Result<()> {
    // parse 32-byte header, validate magic + version
    // walk TOC: numStreams × [u64 compressedSize][u64 uncompressedSize]
    for (offset, size) in &compressed_offsets {
        let compressed = &self.raw[*offset..*offset + *size];
        let mut decoder = ruzstd::StreamingDecoder::new(compressed)?;
        decoder.read_to_end(&mut self.buffer)?;
    }
    self.init_state(version, num_splats, sh_degree, fractional_bits, flags)?;
    self.poll_sections()?;
    self.done = true;
    Ok(())
}
```

**Quaternion branch** extended to `version >= 3` (was `== 3`), since v4 uses the same smallest-three encoding as v3.

The v1–v3 gzip path is **byte-for-byte unchanged**.

lukeyreyno:
Would it be preferable to use the npm spz package?

https://www.npmjs.com/package/@adobe/spz

udwinj:
Thanks for the feedback @lukeyreyno. I looked at `@adobe/spz` carefully and concluded it isn't the right fit for this PR. 

Two reasons:
### 1. It can't replace the Rust WASM path (the main viewer load path)

Spark has **two** SPZ decoders, one is in Rust:

```
SplatLoader → workerPool → worker.ts → rust/spark-lib/src/spz.rs (compiled to WASM)
```

This Rust decoder is baked into `spark-worker-rs` and runs inside the splat-loading worker. `@adobe/spz` is a JS+WASM package (~820 KB) designed to be called from the main thread / JS context. It can't be swapped into the Rust pipeline without gutting the worker architecture and shipping a second WASM module inside the worker. So the Rust-side v4 support added in this PR is necessary regardless of what we do on the TS side.

### 2. It would regress LOD support on the TS path

The TypeScript `SpzReader` / `SpzWriter` (used by `oldWorker.ts` and `transcodeSpz`) supports spark's `flagLod` extension, which is a `numSplats × 2` byte `childCounts` block plus a `numSplats × 4` byte `childStarts` block appended to the splat data inside the gzip stream of v1–v3 files. `oldWorker.ts` actively reads these via the `childCounts` / `childStarts` callbacks when `spz.flagLod` is set:

```ts
if (spz.flagLod) {
  const childCounts = new Uint16Array(numSplats);
  const childStarts = new Uint32Array(numSplats);
  // ... wired into parseSplats callbacks
}
```

`flagLod` is **spark-specific**, not part of the official SPZ spec, so `@adobe/spz` doesn't expose this trailer. Swapping the TS reader to `@adobe/spz` would silently drop child-count/child-start data for any LOD-encoded SPZ that goes through `oldWorker`. Working around it (gunzip ourselves with fflate, extract the trailer at a known offset, re-gzip a flag-cleared buffer before handing to `@adobe/spz`) is possible but adds a gunzip+regzip round-trip and a parallel code path. At that point we've kept most of the complexity and given up the simplification the swap was supposed to buy.

asundqui:
@dmarcos @mrxz  could either of you also give this a look-through? Nothing stuck out to me... In principle I think it's great to add this support to Spark! I see there's both a Rust and TS implementation, nice. I almost feel like we should have a validation test set for file types + versions :)

rust/spark-lib/src/spz.rs
```diff
@@ -39,6 +56,8 @@ impl<T: SplatReceiver> SpzDecoder<T> {
             gzip_header_done: false,
             out_pos: 0,
             done: false,
+            raw: Vec::new(),
+            v4_decoded: false,
```
mrxz: This flag can be merged with the `done` flag.

rust/spark-lib/src/spz.rs
```diff
@@ -97,6 +128,91 @@ impl<T: SplatReceiver> SpzDecoder<T> {
...
+                toc_byte_offset, NGSP_HEADER_SIZE
+            ));
+        }
+        let toc_size = num_streams.checked_mul(TOC_ENTRY_SIZE)
+            .ok_or_else(|| anyhow::anyhow!("v4 TOC size overflow"))?;
+        let toc_end = toc_byte_offset.checked_add(toc_size)
+            .ok_or_else(|| anyhow::anyhow!("v4 TOC end overflow"))?;
+        if self.raw.len() < toc_end {
+            return Ok(()); // need more bytes
+        }
```
mrxz: There are now three points in this function that "need more bytes", the header, the table of contents and the compressed streams. Since they happen in sequence, the header and TOC are now read repeatedly while waiting on bytes from the compressed streams.

Similar to the `poll_sections` this could be handled as a state machine, going through thee above three parts (header, TOC, compressed streams).

rust/spark-lib/src/spz.rs
```diff
@@ -97,6 +128,91 @@ impl<T: SplatReceiver> SpzDecoder<T> {
...
+        let version = read_u32_le(&self.raw[4..8]);
+        if version != 4 {
+            return Err(anyhow::anyhow!("Unsupported NGSP version: {}", version));
+        }
+        let num_splats = read_u32_le(&self.raw[8..12]) as usize;
+        let sh_degree = self.raw[12] as usize;
+        let fractional_bits = self.raw[13];
+        let flags = self.raw[14];
+        let num_streams = self.raw[15] as usize;
+        let toc_byte_offset = read_u32_le(&self.raw[16..20]) as usize;
```
mrxz: Since the v4 header is building upon the header format from v1-3, it should be doable to extend and re-use the pre-existing `poll_header` method reducing code duplication. The `num_streams` and `toc_byte_offset` can be added to the state.

rust/spark-lib/src/spz.rs
```diff
@@ -551,14 +667,65 @@ fn parse_gzip_header(buffer: &mut Vec<u8>) -> anyhow::Result<bool> {
...
+            self.raw.extend_from_slice(bytes);
+            if self.raw.len() < 4 {
+                return Ok(());
+            }
+            let magic = read_u32_le(&self.raw[0..4]);
+            if magic == SPZ_MAGIC {
+                self.format = SpzFormat::Ngsp;
+                // Try to decode if we already have enough bytes.
+                self.try_decode_v4()?;
+                return Ok(());
```
mrxz: Instead of calling `try_decode_v4` (and `poll_decompress` in the GZIP path) here, it'd be cleaner if this `SpzFormat::Unknown` block would focus on format detection, set `self.format` and simply fall-through to the `match self.format` code below.

rust/spark-lib/src/spz.rs
```diff
@@ -551,14 +667,65 @@ fn parse_gzip_header(buffer: &mut Vec<u8>) -> anyhow::Result<bool> {
...
+        match self.format {
+            SpzFormat::Gzip => {
+                self.poll_decompress()?;
+                if !self.done {
+                    return Err(anyhow::anyhow!("Truncated gzip stream"));
+                }
+            }
+            SpzFormat::Ngsp => {
+                // Force a decode attempt; will error if file is truncated.
+                self.try_decode_v4()?;
```
mrxz: What is the purpose of this method call? AFAICT between the last `push` and this `finish` there isn't going to be any new data to try and decode/parse. Either the stream is incomplete or the decoder stage isn't in a terminal state, both cases will already be handled.

mrxz:
> #### New dependency — `ruzstd 0.7.3`
> 
> Pure-Rust ZSTD decoder, no C bindings, compiles cleanly to WASM.

Since I'm not too familiar with the Rust ecosystem, why prefer `ruzstd` over the `zstd` crate? Does `zstd` not compile to WASM? From what I gather from `ruzstd`'s [own description](https://github.com/KillingSpark/zstd-rs#decompression) it's still somewhat work in progress and hasn't reached performance parity yet (emphasis mine):

> Measuring with the 'time' utility the original zstd and my decoder both decoding the same enwik9.zst file from a ramfs, **my decoder is about 3.5 times slower**. Enwik9 is highly compressible, for less compressible data (like a ubuntu installation .iso) my decoder comes close to only **being 1.4 times slower**.

Of course the performance characteristics might be different when compiled to WASM, but if there is a clear speed benefit to the `zstd` crate, I'd rather go for that.

---
SPZv4 has added support for 4th degree spherical harmonics. Spark does not support this, so these files should either be rejected or handled in such a way that the 0-3 SH coefficients are extracted from the stream. Haven't tested it myself with this PR (don't have a sample .spz file with SH degree 4 at hand), but based on the code I believe it'll go wrong at the moment.

As @asundqui mentions, ideally we'd have a set of validation files. The spz repo does have sample files, but these don't appear to be v4. I did see this issue on their repo: https://github.com/nianticlabs/spz/issues/87. If such files are added in the future we might be able to use them.

rust/spark-lib/src/spz.rs
```diff
@@ -97,6 +128,91 @@ impl<T: SplatReceiver> SpzDecoder<T> {
...
+            return Err(anyhow::anyhow!("Invalid v4 SPZ magic: 0x{:08x}", magic));
+        }
+        let version = read_u32_le(&self.raw[4..8]);
+        if version != 4 {
+            return Err(anyhow::anyhow!("Unsupported NGSP version: {}", version));
+        }
+        let num_splats = read_u32_le(&self.raw[8..12]) as usize;
+        let sh_degree = self.raw[12] as usize;
+        let fractional_bits = self.raw[13];
+        let flags = self.raw[14];
```
mrxz: Similar to the official spz library, we should detect if extensions are used for the file and log a warning if so, see https://github.com/nianticlabs/spz/blob/7ae1621e54e4b42c3c9c192b366d09116e558e19/src/cc/load-spz.cc#L676-L680

rust/spark-lib/src/spz.rs
udwinj: Addressed by merging the v4_decoded state into the existing `done` flow: https://github.com/sparkjsdev/spark/pull/332/changes/c42cb9f82aa6817561f8cb88e6e0c6eec8ad32c8

rust/spark-lib/src/spz.rs
udwinj: Addressed in https://github.com/sparkjsdev/spark/pull/332/changes/5faeb6bc58ba0ebff4d1f985cbc98fd8963cc9a2 by restructuring the v4 decoder into an explicit staged state machine: 

`try_decode_v4()` now progresses through `NeedHeader -> NeedToc -> NeedStreams -> Done`, carrying parsed state forward between `push()` calls instead of reparsing the header and rewalking the TOC while waiting on stream bytes.

rust/spark-lib/src/spz.rs
udwinj: https://github.com/sparkjsdev/spark/pull/332/changes/37fdc4bc39a58054953209557266344e15ebd010 - Extracted the shared 15-byte SPZ header prefix parsing into a common `parse_common_header()` helper that is now used by both the legacy gzip `poll_header()` path and the v4 `parse_v4_header()` flow.

That consolidates the shared magic/version/metadata parsing while still letting each path own its version validation and any format-specific fields (`num_streams`, `toc_byte_offset`, gzip-specific handling, etc.) beyond the common prefix.

I'd prefer not to add `num_streams` and `toc_byte_offset` to the shared state. They're v4-internal plumbing used only by `walk_v4_toc` to compute compressed stream offsets, then never referenced again. The downstream section state machine (`SpzDecoderState`) doesn't need them to decode Centers/Alphas/Rgb/Scales/Quats/SH.

Putting them on `SpzDecoderState` would add two permanently`None`/zero fields on every v1 to v3 decode and logically couple "where the v4 file lays out its streams" to "how to decode splat bytes" without functional benefit. Both fields currently live on `V4HeaderInfo` (inside `V4Stage::NeedToc` / `V4Stage::NeedStreams`), which matches their natural lifetime. 

Happy to revisit if you feel strongly, or if there is something I'm missing.

rust/spark-lib/src/spz.rs
udwinj: https://github.com/sparkjsdev/spark/pull/332/changes/b836655b6ca55b9d256e32fc62636d63de1e37c9 - Restructured `push()` into explicit append/advance phases per your suggestion. The `Unknown` block now only handles format detection and buffer placement, then falls through to a single dispatch match.

rust/spark-lib/src/spz.rs
udwinj: Removed as this is a no-op: https://github.com/sparkjsdev/spark/pull/332/changes/ecc2ae50a59cd00ddd97660a5b9dd8e4bcf6da3d

rust/spark-lib/src/spz.rs
udwinj: Done in - https://github.com/sparkjsdev/spark/pull/332/changes/a6ed5936a0192b1c3a8a8219cbe2ef7083944833

udwinj:
> > #### New dependency — `ruzstd 0.7.3`
> > Pure-Rust ZSTD decoder, no C bindings, compiles cleanly to WASM.
> 
> Since I'm not too familiar with the Rust ecosystem, why prefer `ruzstd` over the `zstd` crate? Does `zstd` not compile to WASM? From what I gather from `ruzstd`'s [own description](https://github.com/KillingSpark/zstd-rs#decompression) it's still somewhat work in progress and hasn't reached performance parity yet (emphasis mine):
> 
> > Measuring with the 'time' utility the original zstd and my decoder both decoding the same enwik9.zst file from a ramfs, **my decoder is about 3.5 times slower**. Enwik9 is highly compressible, for less compressible data (like a ubuntu installation .iso) my decoder comes close to only **being 1.4 times slower**.
> 
> Of course the performance characteristics might be different when compiled to WASM, but if there is a clear speed benefit to the `zstd` crate, I'd rather go for that.
> 
> SPZv4 has added support for 4th degree spherical harmonics. Spark does not support this, so these files should either be rejected or handled in such a way that the 0-3 SH coefficients are extracted from the stream. Haven't tested it myself with this PR (don't have a sample .spz file with SH degree 4 at hand), but based on the code I believe it'll go wrong at the moment.
> 
> As @asundqui mentions, ideally we'd have a set of validation files. The spz repo does have sample files, but these don't appear to be v4. I did see this issue on their repo: [nianticlabs/spz#87](https://github.com/nianticlabs/spz/issues/87). If such files are added in the future we might be able to use them.

**ruzstd vs zstd**

@mrxz could use your input here. I originally went with `ruzstd` because of toolchain simplicity: it's pure Rust and compiles cleanly to `wasm32-unknown-unknown` with no additional build dependencies. `rustup target add wasm32-unknown-unknown` is all that's needed. The `zstd` crate's `libzstd-sys` calls a C compiler in its build script and for the `wasm32-unknown-unknown` target requires a clang that targets wasm32 (either wasi-sdk or Emscripten), which isn't currently set up in spark's `rust/build_wasm.js` / wasm-pack flow.

Are we okay with switching to the `zstd` crate (which will make the toolchain more complex but likely speed up SPZ?). The proposed toolchain changes:

- Add wasi-sdk or Emscripten as a build prerequisite documented in the repo
- Update `rust/build_wasm.js` / CI to make the C-to-wasm32 compiler available before invoking wasm-pack

**SH degree 4**

This is already handled in the Rust path. `SpzDecoderState::new` rejects `sh_degree > 3` with `anyhow::anyhow!("Invalid SH degree: {}", sh_degree)`, and `init_state` calls into that on the v4 path before any SH bytes are touched. So a v4 file with `sh_degree == 4` fails fast with an error.

I'll improve the error message to be more specific about v4/SH degree 4 (something like "SPZ SH degree 4 is not supported by this decoder (handles 0–3)").

Extracting degrees 0–3 from a degree-4 file by consuming and discarding the degree-4 bytes in the section state machine is doable but non-trivial and not something I'd want to bundle into this PR. Happy to follow up with a separate PR for that if there's appetite.

mrxz:
> Are we okay with switching to the zstd crate (which will make the toolchain more complex but likely speed up SPZ?). 

It might be easier to go for ruzstd for now. While I do think the performance gain can be worth the possible complexity, without measuring/quantifying it first, it's not really possible to make that call. It might very well be that the speed difference isn't there when compiled to WASM, or that decompressing is only a fraction of the SPZ load time (i.e. only marginal gains possible). 

Having SPZ v4 support in the first place is worth it, and the implementation can always be changed without impacting users.

> Extracting degrees 0–3 from a degree-4 file by consuming and discarding the degree-4 bytes in the section state machine is doable but non-trivial and not something I'd want to bundle into this PR. Happy to follow up with a separate PR for that if there's appetite.

Agreed, better left for a future PR. In practice it's going to be unlikely people will use degree-4 files with Spark. As long as Spark doesn't support it in terms of rendering, you'd only be loading larger-than-need-be files with no visual uplift.

@asundqui @dmarcos Do you guys agree as well?

udwinj:
@mrxz - I updated the error message about v4/SH degree 4 with https://github.com/sparkjsdev/spark/pull/332/changes/5a05962a440872da636b0d770a4c7f9633d94af7, and resolved lint errors with https://github.com/sparkjsdev/spark/pull/332/changes/3fb5893154d43e4979b5719a1f046ecdc3628fcb. 

 This should be all set to merge, unless @asundqui @dmarcos have any additional feedback.

udwinj:
Hi @mrxz - just checking in to see if you need anything else on this PR, or if we are clear to merge. Appreciate your help!

mrxz:
> Hi @mrxz - just checking in to see if you need anything else on this PR, or if we are clear to merge. Appreciate your help!

No pending remarks from my side. Would still like to have input from @asundqui and @dmarcos. Though otherwise I think it's pretty much good to go. Further improvements can always be done down the line, while having SPZv4 support is nice.

That said, I have just opened #374 as keeping around both JS and Rust based decoders for each format isn't really practical. In the context of this PR that could mean removing the JS based changes, so that this PR focusses entirely on the Rust side of things. That would also save us from adding the `@bokuweb/zstd-wasm` dependency to the bundle.

dmarcos:
looks good to me

chipironcin:
@udwinj may I help with removing the JS based changes, as per @mrxz suggests to move this forward?

mrxz:
Now that #374 has been merged, the JS based changes are no longer relevant and can be removed. Other than that, no further pending remarks from my side. Would be nice to get this one finally merged. @udwinj let me know if you have the bandwidth to update/rebase this PR, otherwise I could pick it up from here if you'd like.

asundqui:
I'm sorry for my hiatus, this would be a great feature to merge into Spark, even with the caveats (ruzstd, SH4).

I have also grappled with ruzstd vs. zstd library in the past, and have experienced issues getting the Rust zstd compiled and used with other (non-Cargo) build systems, so for simplicity I agree it would be fine to use ruzstd for now. Perhaps what we could do in the future is add a Rust feature flag that selects the zstd crate instead, but make it optional? I haven't explored if this is possible but maybe having the option for people who really want decompression speed would be nice.

Re: SH4 I think throwing an error for now is a good idea - otherwise people may not know that their SH4 is not being used and would just see worse results.

I guess the final thing left to do is remove the JS changes? @udwinj if you still have interest in this PR we would love to get it across the finish line. If you're not available we could also have @mrxz  or @chipironcin who volunteered to help with this. cc: @dmarcos @oscarlorentzon

dmarcos:
this needs rebase

PR #335, opened 2026-05-06
Combine spark-worker-rs and spark-rs into one

mrxz:
Related to #321 and builds upon: #322 

This PR combines the `spark-worker-rs` and `spark-rs` projects into one, reducing bloat of shared code/logic that would otherwise end up in both `.wasm` files. To avoid the combined `.wasm` file from being embedded multiple times into the bundle, the WASM module is now passed along to the workers, removing the embedded copy in the embedded worker code.

The flow is as follows:

- `wasm.ts` automatically kicks off the WASM compiling
- `wasm.ts` automatically kicks off the initialization of the `spark-rs` project for the main thread
- The `SplatMesh` no longer statically initializes `spark-rs`, but instead checks the initialization status through `wasm.ts`
- The `SplatPager` no longer initializes `spark-rs`, but instead awaits initialization through `wasm.ts`
- The workers now await an `init-wasm` message with the WASM module

This has a couple of benefits, namely that the wasm blob is only embedded once. The [webpack/Next.js workaround](https://github.com/sparkjsdev/spark/pull/231) is no longer needed. And the WASM module is compiled only once across main thread and workers.

With one less rust project, the `build:wasm` script also takes less time. :-)


dmarcos:
Sounds reasonable to me. Thanks!

asundqui:
I fully support this and the earlier #322 ! Big changes and improvements, thanks.

PR #336, opened 2026-05-06
Externalize three/addons from bundle

mrxz:
Imports from `three/addons` were not treated as external during build, meaning the final bundle included things like `FullScreenQuad`. This PR simply expands the external list with a regex matching anything imported from `three/addons`. Helps a bit with reducing the bundle size (#321)


dmarcos:
Thanks!

Issue #329, opened 2026-05-02
Spark v2 fails to load plain CloudCompare binary PLY point cloud (Missing scale_0 property)


Huciko:
Spark v2 appears to fail when loading a plain x/y/z/r/g/b CloudCompare-exported binary PLY point cloud.

The same type of file was loading successfully with the previous Spark version I was using, @sparkjsdev/spark@0.1.10. After upgrading to @sparkjsdev/spark@2.0.0, Spark now appears to treat this plain point-cloud PLY as a Gaussian Splat PLY and expects Gaussian-specific properties such as scale_0.

This seems inconsistent with the Spark v2 documentation, which says .ply loading supports original gsplat PLY, compressed SuperSplat/gsplat PLY variants, and plain x/y/z/r/g/b point clouds, with .ply / .spz auto-detected from file contents.

Docs reference: https://sparkjs.dev/docs/loading-splats

**Environment**
@sparkjsdev/spark: 2.0.0
three: tested in my app with 0.184.0
App: Vite + TypeScript + Three.js

**PLY file details**
The file is a CloudCompare-exported point-cloud PLY.

**Original header:**
ply
format binary_little_endian 1.0
comment Created by CloudCompare v2.13.0 (Kharkiv - Feb 14 2024)
comment Created 2024-03-29T00:12:28
obj_info Generated by CloudCompare!
element vertex 279910
property float x
property float y
property float z
property uchar red
property uchar green
property uchar blue
end_header

**There are no Gaussian Splat properties such as:**
scale_0
scale_1
scale_2
rot_0
rot_1
rot_2
rot_3
opacity

This is expected because it is a plain point-cloud PLY with positions and RGB colors.

What I tried
1. Direct SplatMesh URL load, no fileType
const tree = new SplatMesh({
    url: './images/tree.ply',
});

scene.add(tree);

Result:

Invalid PLY file
2. Removed CloudCompare obj_info line

Because Spark first complained about:

Unsupported PLY header line: obj_info Generated by CloudCompare!

I created a copy of the binary PLY with only this header line removed:

obj_info Generated by CloudCompare!

The binary vertex body was left untouched.

Then I tested:

const tree = new SplatMesh({
    url: './images/Tree-no-obj-info.ply',
});

scene.add(tree);

Result:

Missing scale_0 property
3. Direct SplatLoader, no explicit file type
const loader = new SplatLoader();

const loadedSplats = await loader.loadAsync('./images/Tree-no-obj-info.ply');

[...]

PR #338, opened 2026-05-08
Ignore obj_info header lines in PLY files

mrxz:
The `PlyDecoder` errors upon encountering an unknown/unsupported line in the header. This includes `obj_info`, which servers a similar purpose to `comment`. This PR simply treats "obj_info" the same way as "comment" ensuring .ply files with such lines can be loaded.

See #329 for additional context.


dmarcos:
Thanks!

PR #340, opened 2026-05-11
Support plain x/y/z/r/g/b point clouds in ply files

### LINKED ISSUE #329: Spark v2 fails to load plain CloudCompare binary PLY point cloud (Missing scale_0 property)

mrxz:
Fixes #329 

This PR implements point cloud decoding support in the `spark-lib` project. It only considers a .ply file a point cloud if it contains all of the following properties: `x`, `y`, `z`, `red`, `green`, `blue`. It also supports an optional `alpha` property, similar to how Spark 0.1 handled it.

Since point clouds don't have scale or rotation, a fixed scale of 0.001 is chosen for all three axis. This is the same default/fallback value as in Spark 0.1, but unlike the old approach, this value can't be configured as the constant is on the Rust side of the project.

It's likely that there are still differences in behaviour between this implementation and the old one, especially given that the old one mixed splat and point cloud handling (e.g. it was theoretically possible to have point clouds with spherical harmonics). Testing was done by loading a coloured point cloud file exported from [CloudCompare](https://www.cloudcompare.org)

### DIFF
--- a/rust/spark-lib/src/ply.rs
+++ b/rust/spark-lib/src/ply.rs
@@ -10,6 +10,8 @@ pub const PLY_MAGIC: u32 = 0x00796c70; // "ply"
 const MAX_SPLAT_CHUNK: usize = 65536;
 const SH_C0: f32 = 0.28209479177387814;
 const SUPER_CHUNK_SIZE: usize = 256;
+const POINT_CLOUD_PROPERTIES: [&str; 6] = ["x", "y", "z", "red", "green", "blue"];
+const DEFAULT_POINT_SCALE: f32 = 0.001;
 
 pub struct PlyDecoder<T: SplatReceiver> {
     splats: T,
@@ -69,6 +71,14 @@ impl<T: SplatReceiver> PlyDecoder<T> {
                 lod_tree: false,
             })?;
             PlyState::SuperSplat(state)
+        } else if parsed.is_pointcloud {
+            let state = PointCloudDecoderState::new(parsed.num_splats, parsed.vertex.record_size, parsed.vertex.properties.clone())?;
+            self.splats.init_splats(&SplatInit {
+                num_splats: parsed.num_splats,
+                max_sh_degree: 0,
+                lod_tree: false,
+            })?;
+            PlyState::PointCloud(state)
         } else {
             let state = PlyDecoderState::new(parsed.num_splats, parsed.vertex.record_size, parsed.vertex.properties.clone())?;
             self.splats.init_splats(&SplatInit {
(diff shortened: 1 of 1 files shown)


asundqui:
Love it, thank you for bringing this feature back @mrxz . @dmarcos  we should merge this right?

dmarcos:
Thanks @mrxz

PR #344, opened 2026-05-14
Add experimental SparkRenderer.lodTraverseMode=dynamic/standard for fast approximate LoD traversal mode switching

asundqui:
Hi @dmarcos @mrxz  I have a draft of an alternate LoD traversal algorithm I experimented with a little while back. It removes the "hard limit" on the splat count N, allowing the traversal to run in something closer to O(N) rather than O(N log N). On typical scenes the LoD traversal takes 20-25% of the original time.

Unfortunately, when an LoD traversal completes it updates the splat set that is rendered via texture upload. When this happens too frequently a "judder" becomes quite apparent... so although faster LoD updates are good, the overall experience is not necessarily better. See the video, in the console log you can see the LoD update time when switching between standard and the new dynamic mode.

I was wondering if you guys might be able to help figure out how to make this work judder-free? With this we can update the LoD for large scenes in 100-250ms which I think is a much improved experience.

https://github.com/user-attachments/assets/c0265025-79c7-4035-9300-3ae5c2d42863

### DIFF
--- a/examples/editor/index.html
+++ b/examples/editor/index.html
@@ -560,6 +560,7 @@
     const lodFolder = gui.addFolder("Level-of-Detail").close();
     lodFolder.add(splatExtra, "lod").name("Create Level-of-Detail on load")
     lodFolder.add(spark, "lodSplatScale", 0.001, 3.0, 0.001).name("LoD splat scale").listen();
+    lodFolder.add(spark, "lodRenderScale", 0.1, 10.0, 0.1).name("Min LoD pixel size").listen();
     lodFolder.add(guiOptions, "splatCount").name("Current splat count").listen();
     lodFolder.add(spark, "coneFov0", 0.0, 170.0, 1.0).name("Cone Fov0").listen().onChange(updateConeMeshes);
     lodFolder.add(spark, "coneFov", 0.0, 170.0, 1.0).name("Cone Fov").listen().onChange(updateConeMeshes);
@@ -572,6 +573,7 @@
     lodFolder.add(spark, "behindFoveate", 0.0, 1.0, 0.01).name("Behind Foveate").listen();
     lodFolder.add(spark, "enableDriveLod").name("Enable LoD updates").listen();
     lodFolder.add(spark, "lodInflate").name("Soften LoD splats").listen();
+    lodFolder.add(spark, "lodTraverseMode", ["standard", "dynamic"]).name("LoD traverse mode").listen();
 
     const pageColoring = dyno.dynoBool(false);
     lodFolder.add(pageColoring, "value").name("Page Coloring").listen().onChange(() => {
--- a/rust/spark-rs/src/lod_tree.rs
+++ b/rust/spark-rs/src/lod_tree.rs
@@ -627,3 +627,237 @@ fn compute_pixel_scale<'a>(
     };
     foveate * pixel_scale
 }
+
+#[wasm_bindgen]
+pub fn dynamic_traverse_lod_trees(
+    max_splats: u32, pixel_scale_limit: f32, last_pixel_limit: Option<f32>,
+    // lod_instances: &Array,
+    lod_ids: &[u32], root_pages: &[u32],
+    view_to_objects: &[f32], lod_scales: &[f32],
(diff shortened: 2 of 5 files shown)


mrxz:
Nice improvement in terms of lod traversal time. The judder is unfortunate, though in my testing it wasn't as severe as shown in the video, very infrequent if anything. So it might be worth merging this to already give people to option to opt into it.

Ideally we'd find a way to avoid the judder entirely so this traversal mode can become the default. Though I'm afraid the core of the issue is simply that the texture uploads is synchronous, which is a real annoyance in WebGL. But the common workarounds of throttling/spreading out uploads isn't really usable here, I'm afraid.

There is one theoretical hack I've been thinking of, but haven't tried yet. In WebGL texture uploads can also take other objects as source, not just typed arrays. Things like `ImageBitmap` or `OffscreenCanvas`, which browsers _may_ perform as GPU->GPU copy in case the backing data is already on the GPU. Nothing guarantees this behaviour, though, so on devices that do not (or cannot) optimize this path, it will probably end up being even slower.

It's one of the areas where WebGPU really shines...

asundqui:
@mrxz thanks for your feedback! What's weird with the judder issue when I was trying to fix it is it's almost as if there was a "tripping point" in upload size where the uploads would suddenly cause huge delays (20-30 ms sometimes), whereas uploading fewer splats would be very fast (less than 1 ms). When I did some back-of-the-envelope math for upload bandwidth it seemed like there must been something else funny going on to suddenly start causing huge slowdowns in the upload.

In any case I agree this is an additional option that doesn't take away from the existing algorithm, and I made the default the old algorithm so this shouldn't affect anyone. So maybe we just merge it for now as an option and in the future we can improve on it!

I like the idea of doing the LoD calculation on the GPU. I'm not so keen on WebGPU however, simply because it's not guaranteed to be supported everywhere yet :/. I wonder if there's some way to do a similar LoD calculation with WebGL2...

@dmarcos  should we merge this in? I tested this and everything seems to work. I also added an option to examples/editor to switch the LoD traversal mode if people want to try it.

Finally, this PR also includes a fix to paged splats not being cleaned up properly if removed.

mrxz:
> When I did some back-of-the-envelope math for upload bandwidth it seemed like there must been something else funny going on to suddenly start causing huge slowdowns in the upload.

With the amount of layers behind the `texSubImage2D` call, it's not unsurprising that bandwidth might not be the main factor here. There's this repo [mvaligursky/webgl-webgpu-texture-upload](https://github.com/mvaligursky/webgl-webgpu-texture-upload) that measures various ways of uploading textures in WebGL and WebGPU. You could run it yourself to see if there is indeed a larger jump for higher resolutions than you'd expect.

It also shows that the results can differ between browser, OS and whether or not the GPU is under load or not. 

> I like the idea of doing the LoD calculation on the GPU. I'm not so keen on WebGPU however, simply because it's not guaranteed to be supported everywhere yet :/. I wonder if there's some way to do a similar LoD calculation with WebGL2...

WebGL2 and WASM isn't guaranteed to be supported everywhere either... Of course, _requiring_ WebGPU at this point is likely going to be too restrictive, but I do think there's value in preparing for and even supporting WebGPU at this point.

For example, the LoD calculations could be done in a separate WebGPU context. Using an `OffscreenCanvas` the result could then be uploaded into the WebGL texture. Enabling both GPU acceleration and avoiding a GPU->CPU->GPU round trip. On systems that don't support WebGPU the current approach could still be used as a fallback.

isaac-mason:
> WebGL2 and WASM isn't guaranteed to be supported everywhere either... Of course, requiring WebGPU at this point is likely going to be too restrictive, but I do think there's value in preparing for and even supporting WebGPU at this point.

Also worth mentioning that THREE.WebGPURenderer has a WebGL fallback for this reason. Although I imagine it would be limiting to rely on [TSL's](https://github.com/mrdoob/three.js/wiki/Three.js-Shading-Language) auto-WebGL fallback capabilities entirely, probably large chunks of functionality could be written once with TSL to support WebGPU and WebGL.

See also: https://github.com/mrdoob/three.js/pull/32851

Aside, spark playing nice with three's TSL would be very interesting (this role currently being played by dyno).

PR #349, opened 2026-05-18
Include three/addons/ in importmaps and switch to jsdelivr CDN

mrxz:
With #336 the bundle expects to be able to resolve imports from `three/addons/`. However, not all examples included `three/addons/` in their importmap, causing them to be broken on the deployed website.

This PR updates all importmaps across examples and documentation to ensure they include both `three` and `three/addons/` as per the [installation instructions](https://threejs.org/manual/#en/installation) of Three.js. Additionally, since cdnjs does not expose the `examples` folder of Three.js, the CDN was switched to jsdelivr. Users would have to do this either way as soon as they wanted to use any of the Three.js addons themselves.

Note: an alternative solution would be to vendor the `Pass.js` file from Three.js so that the spark bundle doesn't import from `three/addons/` directly.

@dmarcos The current process for docs and examples feels brittle. When testing changes to the examples locally, Vite handles the resolving and ignores the importmap, hiding any issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,8 @@ Copy the following code into an `index.html` file.
 <script type="importmap">
   {
     "imports": {
-      "three": "https://cdnjs.cloudflare.com/ajax/libs/three.js/0.180.0/three.module.js",
+      "three": "https://cdn.jsdelivr.net/npm/three@0.180.0/build/three.module.js",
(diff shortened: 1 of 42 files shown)


dmarcos:
thanks!

PR #354, opened 2026-05-20
Improve build experience with regards to Rust Wasm components

mrxz:
Related to #195 and #304.

Removes `npm run build:wasm` from the `prepare` script to allow people to install the pre-built bundles from the repository directly without having to have the Rust toolchain setup on their machines.

Additionally the PR explicitly adds the `npm run build:wasm` command to the instructions in the README and adds a check to `vite.config.ts` to warn users when they attempt to run or build the project without first building the Rust Wasm components. Additionally I updated the ignore list in `biome.json` to ignore the `rust/` folder in its entirety (see https://github.com/sparkjsdev/spark/pull/350#issuecomment-4480536794).

Example of the error message:
image


dmarcos:
Thank you!

PR #356, opened 2026-05-21
Render gaussian splats relative to the camera independent of (large) global coordinates

pmi-gbs:
I've had issues with big global coordinates (which happens e.g. when using WGS84 cartesian CRS). This happened within the discussion in https://github.com/WilliamLiu-1997/3D-Tiles-RendererJS-3DGS-Plugin/issues/2 where even parenting the Renderer to the camera didn't help.

I think we can generally fix these floating point quantization issues by having the renderer really relative to the camera automatically (so that all the math is done in f64, instead of what I suspect f32 in the shader currently, but I haven't carefully checked this TBH).

So this should get rid of all these issues (including within the old renderer I think), so that the user doesn't have to care anymore about these float quantization artifacts.

As I'm maintaining Open source projects myself I want to disclaim this was heavily assisted by AI to be fair, I tried my best to understand the changes, and avoid slop. But well I'm not (yet) deep into the codebase, and I think the changes are compact enough to be worth putting into a PR. I did instruct that it should avoid allocations, so this resulted in a few drive-by fixes, which I think are fine (in case you're wondering).

### DIFF
--- a/docs/docs/packed-splats.md
+++ b/docs/docs/packed-splats.md
@@ -127,7 +127,7 @@ Opacity is encoded on a linear scale where 0..255 maps to 0..1.
 
 ### Splat center encoding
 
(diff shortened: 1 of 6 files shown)


joaomelorodrigues:
If this works, it will solve a lot of issues

asundqui:
@pmi-gbs this reply is obviously quite late, but I believe this was done before `ExtSplats` was introduced into Spark 2.0. I believe ExtSplats should solve the quantization issues you were seeing, which was from `PackedSplats` storing centers as `fp16`. When enabling `extSplats` in `SparkRenderer` we use `fp32` encoding, and the accumulator is also camera-relative, which I believe your PR was attempting to do. Unfortunately `fp64` is not an option, because although Javascript is natively `fp64` for numbers, consumer-grade GPUs can't do `fp64` efficiently, so we don't ever want that in shaders.

I will be closing this PR, thank you for your contribution, and sorry for your trouble and need for workarounds. cc: @oscarlorentzon @dmarcos

PR #358, opened 2026-06-01
Set dataReady false for textures arrays in SplatPager

mrxz:
The `SplatPager` initializes several texture arrays with newly created (zeroed) arrays. The default behaviour of Three.js is to upload this data to the GPU. In this case, however, there is no need as the data should never be accessed and the GPU initialized texture would already contain zero's either way. This PR sets the `dataReady` flag to false, preventing Three.js from doing the initial upload. When a page needs to be uploaded, only the relevant layers are uploaded, so the superfluous upload is completely mitigated, not postponed.

While making the above change, I got annoyed by the excessive duplication in `SplatPager`. Resetting the flag needed to be repeated several times (once per SH texture x 2 for `extSplats` and non `extSplats`). As such, I've also replaced the individual sh-texture properties with an array, allowing creation, uploading and disposing to be handle by simple for-loops instead.

### DIFF
--- a/src/SplatPager.ts
+++ b/src/SplatPager.ts
@@ -15,7 +15,7 @@ import {
   type SplatEncoding,
   SplatFileType,
 } from "./defines";
-import { pagedSplatTexCoord } from "./dyno";
+import { type DynoUsampler2DArray, pagedSplatTexCoord } from "./dyno";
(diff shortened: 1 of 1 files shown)


dmarcos:
This looks good to me. @asundqui thoughts?

asundqui:
Great simplification and improvement, thank you, let's merge it in!

PR #359, opened 2026-06-01
Fix build-lod SH degree clamping logic

### LINKED ISSUE #352: build-lod panics on SOG inputs when --max-sh exceeds bundled meta.shn.bands
## Summary

`build-lod --max-sh=N` panics with `index out of bounds` when `N` is greater than the SH bands the input file actually contains. The CLI flag behaves as both a lower and upper bound; based on its name it should be a cap only.

This consistently reproduces on `.sog` inputs that bundle no spherical harmonics (`meta.shn` absent or `meta.shn.bands = 0`), which is common for generator-produced SOGs (e.g. World Labs ``spz_urls.full_res`` re-encoded to SOG via ``@playcanvas/splat-transform``).

## Repro

\`\`\`
build-lod --quality --rad-chunked --cluster-sh=5 --max-sh=3 input.sog
\`\`\`

with a SOG whose ``meta.json`` has no ``shn`` section.

## Observed

\`\`\`
Detected file type: SOGS
Read: num_splats: 29427170 with sh_degree: 0
bhatt_lod::compute_lod_tree: initial_len=29427170
Sorted and prepared splats
Level: -8, step: 0.00390625, frontier: 1155 / 29427170, # active: 1155, # splats: 29427170
thread 'main' panicked at spark-lib/src/gsplat.rs:414:36:
index out of bounds: the len is 0 but the index is 1
\`\`\`

## Root cause

In ``rust/spark-lib/src/sogs.rs`` ``decode_v2``, ``max_sh_degree`` is derived from ``meta.shn.bands``; when ``shn`` is missing it defaults to ``0``, so ``sh1``/``sh2``/``sh3`` are initialized as empty ``Vec``s (``rust/spark-lib/src/sogs.rs#L189-L212``).

The CLI then calls ``splats.set_max_sh_degree(max_sh)`` in ``rust/build-lod/src/main.rs#L161-L163`` with the requested CLI value (3). ``TsplatArray::set_max_sh_degree`` (``rust/spark-lib/src/gsplat.rs#L256-L269``) only **truncates** ``sh1/sh2/sh3`` when the new value is lower than the current one. It never grows them, but it does raise ``self.max_sh_degree``.

Subsequently, ``bhatt_lod`` walks the tree and unconditionally accesses ``&self.sh1[index]`` whenever ``self.max_sh_degree >= 1`` (``rust/spark-lib/src/gsplat.rs#L410-L420``). With ``sh1.len() == 0`` and ``max_sh_degree == 3``, the first non-leaf merge panics.

## Suggested fix

Clamp ``--max-sh`` to the source's actual SH degree before applying it. Minimal change in ``rust/build-lod/src/main.rs`` around L161:

\`\`\`rust
if let Some(max_sh) = options.max_sh {
    let input_sh = TsplatArray::max_sh_degree(&splats);
    let clamped = max_sh.min(input_sh);
    splats.set_max_sh_degree(clamped);
    description.insert(\"max_sh_degree\".to_string(), serde_json::Value::Number(clamped.into()));
}
\`\`\`

Alternatively, make ``TsplatArray::set_max_sh_degree`` clamp internally so all callers (including the WASM/in-browser path) are protected:

\`\`\`rust
fn set_max_sh_degree(&mut self, max_sh_degree: usize) {
    assert!(max_sh_degree <= 3, \"SH degrees must be between 0 and 3\");
    let current_capacity = if !self.sh3.is_empty() { 3 }
        else if !self.sh2.is_empty() { 2 }
        else if !self.sh1.is_empty() { 1 }
        else { 0 };
    self.max_sh_degree = max_sh_degree.min(current_capacity);
    // truncation logic stays as-is
}
\`\`\`

## Workaround

Pass ``--max-sh=0`` on SOG inputs known to lack SH, or probe ``meta.shn.bands`` out of the SOG zip and clamp ``--max-sh`` to it before invoking the binary.

## Environment

- ``build-lod`` from the latest ``sparkjsdev/spark`` ``main``
- Input: 29.4M-splat ``.sog`` produced by ``@playcanvas/splat-transform`` from a World Labs ``.spz`` (SH stripped during SOG conversion)
- Host: Ubuntu 22.04 on AWS ``g4ad.xlarge`` (AMD V520, Vulkan/RADV)

asundqui:
Renamed confusing `set_max_sh_degree` to better `clamp_sh_degree` and fix logic. Added the clamping vs. final max_sh_degree to the RAD file annotations.

This fixes #352 . Tested for both Gsplat and Csplat on SOG files with fewer degrees than the max-sh cmdline setting.

### DIFF
--- a/rust/build-lod/src/main.rs
+++ b/rust/build-lod/src/main.rs
@@ -159,8 +159,8 @@ fn process_file_lod_tsplat<TS: SplatReceiver + TsplatArray + SplatGetter>(filena
     }
 
     if let Some(max_sh) = options.max_sh {
-        splats.set_max_sh_degree(max_sh);
-        description.insert("max_sh_degree".to_string(), serde_json::Value::Number(max_sh.into()));
+        splats.clamp_sh_degree(max_sh);
+        description.insert("clamp_sh_degree".to_string(), serde_json::Value::Number(max_sh.into()));
     }
 
     if let Some(min_box) = options.min_box {
@@ -242,7 +242,9 @@ fn process_file_lod_tsplat<TS: SplatReceiver + TsplatArray + SplatGetter>(filena
     let chunk_duration = start_time.elapsed();
     description.insert("chunk_duration".to_string(), serde_json::Number::from_f64(chunk_duration.as_secs_f64()).into());
 
-    let num_sh = TsplatArray::max_sh_degree(&splats).min(options.max_sh.unwrap_or(3));
+    let num_sh = TsplatArray::max_sh_degree(&splats);
+    description.insert("max_sh_degree".to_string(), serde_json::Value::Number(num_sh.into()));
+    
     let mut sh_clusters = None;
     if let Some(num_iterations) = options.cluster_sh {
         if num_sh > 0 {
--- a/rust/spark-lib/src/csplat.rs
+++ b/rust/spark-lib/src/csplat.rs
@@ -163,9 +163,9 @@ impl TsplatArray for CsplatArray {
         self.max_sh_degree
     }
 
-    fn set_max_sh_degree(&mut self, max_sh_degree: usize) {
+    fn clamp_sh_degree(&mut self, max_sh_degree: usize) {
         assert!(max_sh_degree <= 3, "SH degrees must be between 0 and 3");
-        self.max_sh_degree = max_sh_degree;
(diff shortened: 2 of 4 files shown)


mrxz (approved):

mrxz:
LGTM, `clamp_sh_degree` is a lot clearer and the `options.max_sh` argument is only handled once.

PR #361, opened 2026-06-03
Resolve Rust compiler warnings dynamic lod travesal

mrxz:
With the changes in #344 running `npm run build:wasm` logs a few warnings from the Rust compiler. Mainly boiling down to unused variables. This PR resolves them:

- The `_last_pixel_limit` argument is left in to keep the signature identical to the standard `traverse_lod_trees`
- The `performance` object was removed entirely, as the "test time" wasn't measuring anything and the resulting value was unused anyway
- `finer_count` and `limits` have been removed as well, since no consumer read them

Leaving them in (e.g. as commented code) could simplify future testing/development, but this should be avoided. Temporarily adding them back in should be straightforward. If that ends up being a common occurrence/hindrance, using conditional compilation for them would be cleaner IMHO.

### DIFF
--- a/rust/spark-rs/src/lod_tree.rs
+++ b/rust/spark-rs/src/lod_tree.rs
@@ -630,7 +630,7 @@ fn compute_pixel_scale<'a>(
 
 #[wasm_bindgen]
 pub fn dynamic_traverse_lod_trees(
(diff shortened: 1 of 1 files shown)


dmarcos:
Thanks

PR #367, opened 2026-06-08
Extract U32 texture data upload logic and restore pixelStorei parameters

### LINKED ISSUE #366: SparkRenderer leaks WebGL UNPACK_FLIP_Y pixel-store state into the host Three.js renderer → flipped texture re-uploads
**Environment**: 
@sparkjsdev/spark@2.0.0, three@0.184.0, WebGL2, sharing one THREE.WebGLRenderer (Spark added to the scene via scene.add(sparkRenderer)).

**Summary**: 
Spark issues raw gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, …) (and bindFramebuffer/setRenderTarget) directly on the GL context, outside Three's WebGLState tracking, and doesn't restore the host renderer's pixel-store state afterward. Because Three caches pixel-store state and skips redundant pixelStorei calls, Three's cache and the real GL state diverge. The next time Three (re)uploads a texture (e.g. a flipY=true image texture revived after dispose() + needsUpdate=true), Three skips re-applying UNPACK_FLIP_Y and the texture uploads with Spark's leftover state → rendered vertically flipped.

**Observed**:
Only affects texture re-uploads (a disposed-then-revived THREE.Texture); the first upload is fine.
Affects flipY=true textures (standard TextureLoader images); flipY=false GLTF textures are unaffected.
Removing the SparkRenderer from the scene → no flip.
renderer.resetState() before renderer.render(scene, camera) → no flip (forces Three to re-issue the pixel-store flags).

**Minimal repro**:
three scene with a SparkRenderer added to the scene.
Create a mesh with a TextureLoader texture (flipY=true), render → correct.
texture.dispose(), then later reuse the same THREE.Texture with texture.needsUpdate = true and render again.
Texture appears upside-down. Remove SparkRenderer (or renderer.resetState() before render) → correct.
Root cause (confirmed in dist): raw gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, true) outside Three's WebGLState, with no save/restore of host state.

Suggested fixes (any one):

Save and restore the pixel-store flags Spark changes (UNPACK_FLIP_Y_WEBGL, and any others) around Spark's passes, or
Set them via the host renderer's state API / call renderer.resetState() after Spark's passes so Three's cache stays consistent, or
At minimum, document that consumers sharing a Three.js renderer must call renderer.resetState() (or that Spark touches UNPACK_FLIP_Y), and expose an option/hook for state restoration.

mrxz:
Both `SplatPager` and `SparkRenderer` manually uploaded texture data using the `WebGLRenderingContext` directly. This stems from a limitation in Three.js preventing 2D regions from easily being updated, see https://github.com/mrdoob/three.js/issues/32254. This PR extracts the logic into a utility function to avoid duplication.

Additionally the `pixelStorei` parameters set as part of this procedure weren't restored to the values they were before the upload. This becomes a problem with Three.js >= r184 as it now caches the parameter values in its `WebGLState`, see #366. This PR ensures they are now properly restored. Once Spark requires Three.js 184 or later it can switch to `renderer.state.pixelStorei` instead.

Fixes #366

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -20,6 +20,7 @@ import {
   isMobile,
   isOculus,
   isVisionPro,
+  uploadU32DataTextureRows,
 } from "./utils";
 
 export interface SparkRendererOptions {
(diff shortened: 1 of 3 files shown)


dmarcos:
we should probably bump to r184 sooner than later

mrxz:
> we should probably bump to r184 sooner than later

On one hand keeping up with the latest release is nice to ensure it keeps working, but we should stick to the oldest version we intend to support. Especially considering not all frameworks that use Three.js update at the same cadence as Three itself. Unless there is a clear reason for it, Spark shouldn't force users to update Three IMHO.

It's unfortunate when things break/regress like it did here, but that's likely going to be infrequent, whereas building against/relying on new additions/features is a bigger compat risk.

PR #370, opened 2026-06-11
chore(rust): fix clippy warnings in splat sorting functions

mmjinglin163:
chore(rust): fix clippy warnings in splat sorting functions

### DIFF
--- a/rust/spark-rs/src/lib.rs
+++ b/rust/spark-rs/src/lib.rs
@@ -43,7 +43,7 @@ pub fn sort_splats(
 ) -> u32 {
     let max_splats = readback.length() as usize;
 
-    let active_splats = SORT_BUFFERS.with_borrow_mut(|buffers| {
+    SORT_BUFFERS.with_borrow_mut(|buffers| {
         buffers.ensure_size(max_splats);
         let sub_readback = readback.subarray(0, num_splats);
         sub_readback.copy_to(&mut buffers.readback[..num_splats as usize]);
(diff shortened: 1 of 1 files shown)


mrxz:
Thanks for the PR, the changes look good for the most part. Personally I think the elimination of the let binding for `active_splats` reduces readability a bit, but I guess it's more idiomatic.

Though if we're going to use clippy, I'd rather see it integrated and enforced (similar to TS linting through biome). Otherwise we'd end up with having to play catch-up with PRs every time changes to the Rust code introduces new linter warnings. @asundqui any preference in this matter?

mmjinglin163:
Thanks for the review!

Regarding active_splats : I totally agree that readability should come first for maintainers. If you prefer the let binding, I can easily revert that part and add #[allow(clippy::let_and_return)] above it to keep Clippy happy without sacrificing readability. Just let me know if you want me to do that!

Regarding the Clippy integration: I 100% agree. Relying on manual checks is prone to regressions. We should definitely enforce it in CI just like Biome for TS. My preference would be to:

1. Add a cargo clippy --workspace --all-targets --all-features -- -D warnings step to .github/workflows/ci-linux.yml .
2. Add a "lint:rust" script in package.json for local convenience.
I can include this CI setup in this PR, or we can track it in a separate PR if you prefer. What do you think?

asundqui:
@mmjinglin163 thank you  for your PR! @mrxz is this PR superseded by #414 ? If that's the case then can you close this one? I agree with you on readability, prefer leaving this alone. But if we're going full-clippy in #414  then I'm happy to go with whatever the linter says for this.

PR #379, opened 2026-06-23
Use type information of rpcHandlers for `worker.call`

mrxz:
Calling `worker.call` was mostly untyped with `args: unknown` and a return type of `Promise<unknown>` which required all callers to perform type assertions. Generally this meant that both sides needed to be in sync in terms of the expected types. This PR exports the type information of `rpcHandlers` and uses it instead.

Since only the type is exported and imported, this has no impact on the resulting bundle. There were a handful of places that needed minor changes to get the TypeScript compiler to be agree.

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -1046,15 +1046,11 @@ export class SparkRenderer extends THREE.Mesh {
     if (!this.sortWorker) {
       this.sortWorker = new SplatWorker();
     }
-    const result = (await this.sortWorker.call("sortSplats32", {
+    const result = await this.sortWorker.call("sortSplats32", {
       numSplats,
       readback,
       ordering,
-    })) as {
-      readback: Uint32Array<ArrayBuffer>;
-      ordering: Uint32Array;
-      activeSplats: number;
-    };
+    });
 
     if (this.sortDelay > 0) {
       await new Promise((resolve) => setTimeout(resolve, this.sortDelay));
@@ -1250,9 +1246,9 @@ export class SparkRenderer extends THREE.Mesh {
           numFetchers: this.numLodFetchers,
         });
 
-        const { lodId } = (await worker.call("newLodTree", {
+        const { lodId } = await worker.call("newLodTree", {
           capacity: this.pager.maxSplats,
-        })) as { lodId: number };
+        });
         this.pagerId = lodId;
       }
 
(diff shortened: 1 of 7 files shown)


dmarcos:
This needs rebase

mrxz:
@dmarcos Rebased

asundqui:
Yes this is great, very nice cleanup! I like how you impute the return type automatically. Let's merge yes @dmarcos @oscarlorentzon ?

src/worker.ts
```diff
@@ -39,6 +39,7 @@ const rpcHandlers = {
   getLodTreeLevel,
   nextChunk,
 };
+export type rpcHandlers = typeof rpcHandlers;
```
oscarlorentzon, line 42: Do we want `RpcHandlers` for the type? There are no other camelCase exported types in `src/`.

src/SparkRenderer.ts
```diff
@@ -1520,16 +1509,12 @@ export class SparkRenderer extends THREE.Mesh {
...
         maxSplats: Math.min(this.lodRaycast, Math.round(totalLodSplats * 0.1)),
         pixelScaleLimit,
         instances,
-      })) as {
-        keyIndices: Record<
-          string,
-          { lodId: number; numSplats: number; indices: Uint32Array }
-        >;
-      };
+        traverseMode: this.lodTraverseMode,
```
oscarlorentzon, line 1516: The raycast traversal now passes the traverse mode which is a behavioral change. Would it be better to change the behavior in a separate PR if this is what we want?

oscarlorentzon:
> Yes this is great, very nice cleanup! I like how you impute the return type automatically. Let's merge yes @dmarcos @oscarlorentzon ?

Agree, good improvement. 

`npm run build` prints a type error at [`src/SplatLoader.ts:154`](https://github.com/sparkjsdev/spark/blob/85979186baaf40ce4e19f0e0c41253d028673bf6/src/SplatLoader.ts#L154). Main does not so it would be good to fix before we merge.

We might want to add a `typecheck": "tsc -p tsconfig.json --noEmit` script separately to catch this automatically in the future.

src/worker.ts
mrxz: In a similar vein to how `export class A` creates both the value and the type under the same name, I tend to use the same name for the type in cases like this. There doesn't seem to be a precedent in the code base thus far.

I'm fine either way, so if we instead want to go for types always using PascalCase, I can update it to `RpcHandlers` instead.

mrxz:
> `npm run build` prints a type error at [`src/SplatLoader.ts:154`](https://github.com/sparkjsdev/spark/blob/85979186baaf40ce4e19f0e0c41253d028673bf6/src/SplatLoader.ts#L154). Main does not so it would be good to fix before we merge.

Maybe time to update the TypeScript version, as that also resolves the issue. The code already ensures that `fileBytes` is `Uint8Array<ArrayBufferLike> | undefined`, but the older TS compiler can't determine that. 

> We might want to add a typecheck": "tsc -p tsconfig.json --noEmit script separately to catch this automatically in the future.

Good idea, having typechecking as part of the checks that run on PRs would be good to have. @dmarcos

src/worker.ts
oscarlorentzon: Let's keep as is.

PR #380, opened 2026-06-23
Always create a ReadableStream for pushing data into decoder

mrxz:
In the worker the `decodeBytesUrl` might be tasked with loading from `fileBytes`, `url` or `chunked`. This PR changes the three different options to all create a `ReadableStream` allowing them to be treated the same afterwards.

The benefit is that this ensures all three behave comparably, as they now all output progress status events, which the `fileBytes` path didn't before. Additionally it becomes possible to add a `TransformStream` in between regardless of where the data is coming from.

### DIFF
--- a/src/worker.ts
+++ b/src/worker.ts
@@ -96,35 +96,6 @@ function sortSplats32({
   return { activeSplats, readback, ordering };
 }
 
-async function fetchRange({
-  url,
-  requestHeader,
-  withCredentials,
-  offset,
-  bytes,
-}: {
-  url: string;
-  requestHeader?: Record<string, string>;
-  withCredentials?: string;
-  offset?: number;
-  bytes?: number;
-}): Promise<Uint8Array> {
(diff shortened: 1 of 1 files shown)


asundqui:
I love it, this is a great idea, will be nice to be able to see loading/decoding progress for the `fileBytes` path as well. I think we should merge this, agree @dmarcos @oscarlorentzon ?

src/worker.ts
```diff
@@ -144,13 +115,17 @@ async function decodeBytesUrl({
   chunkedLength?: number;
   sendStatus: (data: unknown) => void;
 }) {
+  let readStream: ReadableStream;
```
oscarlorentzon: Should we make this a `ReadableStream<Uint8Array>` to avoid `any` values in the handling?

oscarlorentzon:
Nice unification, one small comment otherwise looks good.

PR #387, opened 2026-07-02
Clamp splat RGB to positive in splatVertex, matching the reference rasterizer

### LINKED ISSUE #386: extSplats: unclamped negative colors become NaN and corrupt float render targets
I recently switched to `extSplats: true` + `accumExtSplats: true` to keep fp16 color precision end to end (picked up from the suggestion in https://github.com/sparkjsdev/spark/issues/383#issuecomment-4832308780 — it does fix the 8-bit banding nicely). My splats are composited through a HalfFloat render target, and with that combination every scene I load comes out like this:

extSplats stock — black blobs and neon speckles

Same scene and camera through the default packed path, and through the ext path with the one-line fix suggested below:

default packed path — reference

extSplats with the fix

A second, unrelated scene shows the same thing:

second scene, extSplats stock

second scene, default packed path

second scene, extSplats with the fix

### What's happening

Trained 3DGS scenes normally contain some negative base colors — 3–8% of splats in my scenes, with no NaN/Inf in the files themselves. The packed path clamps those away at pack time (`packSplatEncoding` in `splatDefines.glsl`), but `packSplatExt` / `packSplatExtCov` pack raw fp16 and nothing downstream clamps either. With `encodeLinear` the negatives then hit `srgbToLinear` — `pow` with a negative base, undefined in GLSL — and come out as NaN on every GPU I tried.

On a normal canvas the fixed-function blend clamps to [0, 1] and hides all of this. In a float render target (any EffectComposer-style pipeline) blending doesn't clamp, so the NaN/negative premultiplied colors land straight in the framebuffer — the black blobs and neon speckles above.

### Repro

Any trained 3DGS .ply with `extSplats: true`, `accumExtSplats: true`, `encodeLinear: true`, rendered into a `HalfFloatType` target and blitted to screen. Drop any one ingredient and it looks fine again — although without `encodeLinear` it's still wrong, just quieter: no NaN, but the negatives darken the blend and leave dark smears in the halo:

encodeLinear off, stock — dark smears in the halo

encodeLinear off, with the fix

### Suggested fix

One line at the top of `packSplatExt` and `packSplatExtCov`, mirroring the clamp the packed encoders already do:

```glsl
rgba = mix(clamp(rgba, 0.0, 65504.0), vec4(0.0), isnan(rgba));
```

Negatives and NaN go to zero (a NaN alpha then falls below `minAlpha` and the splat is discarded), +Inf caps at 65504 — the largest finite fp16 value, i.e. the most `packHalf2x16` can hold — and everything in between passes through, so the HDR headroom of the ext encoding stays intact. With this the ext path matches the 8-bit reference again, and output on a normal canvas is byte-identical before/after. An alternative would be guarding `srgbToLinear` against negative input, but sanitizing at pack time covers every downstream consumer at once.

npm 2.1.0 and current main behave identically. I'll open a PR with this change — would appreciate you taking a look.

viethungle0503:
Fixes #386

Trained 3DGS scenes normally contain some negative base colors. The packed encoders clamp those away at pack time, but `packSplatExt`/`packSplatExtCov` pack raw fp16 — the negatives then turn into NaN in `srgbToLinear` (`pow` with a negative base is undefined in GLSL) and, when rendering into a float target where blending doesn't clamp, end up in the framebuffer as black blobs and neon speckles. Screenshots and the full story in #386.

This adds the same clamp semantics the packed path already has, at the top of both ext pack functions: negatives and NaN go to zero (a NaN alpha then falls below `minAlpha` and the splat is discarded), +Inf caps at fp16 max, and everything in between passes through so the ext encoding keeps its HDR headroom.

Output on a normal canvas is byte-identical before/after — the fixed-function blend clamp was already enforcing these semantics there — so the change only shows up where things were broken.

### DIFF
--- a/src/shaders/splatDefines.glsl
+++ b/src/shaders/splatDefines.glsl
@@ -309,7 +309,6 @@ void packSplatExtCov(
     out uvec4 packedData, out uvec4 packedData2,
     vec3 center, vec4 rgba, vec3 xxyyzz, vec3 xyxzyz
 ) {
-    rgba = mix(clamp(rgba, 0.0, 65504.0), vec4(0.0), isnan(rgba));
     packedData.x = floatBitsToUint(center.x);
     packedData.y = floatBitsToUint(center.y);
     packedData.z = floatBitsToUint(center.z);
@@ -358,7 +357,6 @@ void packSplatExt(
     out uvec4 packedData, out uvec4 packedData2,
     vec3 center, vec3 scales, vec4 quaternion, vec4 rgba
 ) {
-    rgba = mix(clamp(rgba, 0.0, 65504.0), vec4(0.0), isnan(rgba));
     packedData.x = floatBitsToUint(center.x);
     packedData.y = floatBitsToUint(center.y);
     packedData.z = floatBitsToUint(center.z);
--- a/src/shaders/splatDefines.glsl
+++ b/src/shaders/splatDefines.glsl
@@ -309,6 +309,7 @@ void packSplatExtCov(
     out uvec4 packedData, out uvec4 packedData2,
     vec3 center, vec4 rgba, vec3 xxyyzz, vec3 xyxzyz
 ) {
+    rgba = mix(clamp(rgba, 0.0, 65504.0), vec4(0.0), isnan(rgba));
     packedData.x = floatBitsToUint(center.x);
     packedData.y = floatBitsToUint(center.y);
     packedData.z = floatBitsToUint(center.z);
@@ -357,6 +358,7 @@ void packSplatExt(
     out uvec4 packedData, out uvec4 packedData2,
     vec3 center, vec3 scales, vec4 quaternion, vec4 rgba
 ) {
+    rgba = mix(clamp(rgba, 0.0, 65504.0), vec4(0.0), isnan(rgba));
     packedData.x = floatBitsToUint(center.x);
     packedData.y = floatBitsToUint(center.y);
     packedData.z = floatBitsToUint(center.z);
--- a/src/shaders/splatVertex.glsl
+++ b/src/shaders/splatVertex.glsl
@@ -103,6 +103,10 @@ void main() {
         }
     }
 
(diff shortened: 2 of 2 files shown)


mrxz:
Thanks for the pull request, though I don't think `packSplatExt`/`packSplatExtCov` is the right place for this clamping. The `ExtSplats` representation can be used at various places in Spark. When used as splat source for a `SplatMesh` the encoded RGBA values will be _without_ SH contribution, while in the accumulator they will contain the SH contribution. The positive clamping should ideally only be applied in the latter case.

The key difference with `PackedSplats` is that its representation is limited to positive values. There the clamping is out of necessity and not for correctness, though it does have the benefit that it prevents the issue.

In any case, I suggest moving the logic to `splatVertex.glsl` and simplify the logic to only clamp the RGB channels to positive values, matching [graphdeco-inria/diff-gaussian-rasterization `forward.cu#L65-L70`](https://github.com/graphdeco-inria/diff-gaussian-rasterization/blob/59f5f77e3ddbac3ed9db93ec2cfe99ed6c5d121d/cuda_rasterizer/forward.cu#L65-L70). While this wouldn't address the `NaN` or `+Inf` values, I'm not sure 'sanitizing' these is ideal. Unlike negative values, these really shouldn't happen and I'd rather they cause noticeably wrong result, instead of flying under the radar when they do show up.

viethungle0503:
Thanks for the feedback — I hadn't considered that the same encoder also packs the pre-SH source representation, where clamping would shift the SH-evaluated result. I've moved the clamp to `splatVertex.glsl` right after the unpack, as a plain `max(rgba.rgb, vec3(0.0))` matching the rasterizer you linked, and dropped the NaN/+Inf handling — I see the reasoning there, better to let those stay visible than mask them.

Re-ran my scenes to confirm: the float-target corruption is gone, the ext path matches the 8-bit reference again, and output on a normal canvas is still byte-identical before/after.

mrxz:
Looks good to me

Issue #375, opened 2026-06-19
Add support for streamed SOG


jo-chemla:
Quick recap of the context: 
 - sparkjs supports 3dgs scenes encoded as Playcanvas [SOG v2](https://developer.playcanvas.com/user-manual/gaussian-splatting/formats/sog/) format since v0.1.10 as shown on [the doc](https://sparkjs.dev/docs/0.1-2.0-migration-guide/#sogs-file-support) and [this comment](https://github.com/sparkjsdev/spark/issues/86#issuecomment-3585580136) 
 - Sparkjs 2 introduced support for tiled 3dgs scenes in the [RAD format](https://www.worldlabs.ai/blog/spark-2.0#progressive-streaming) as shown in the very detailed v2 blog post, section [level-of-detail](https://www.worldlabs.ai/blog/spark-2.0#level-of-detail). These are ~octree hierarchy with base 1.5, built via `build-lod with bhatt-lod or tiny-lod` modes, where each node contains a rad encoded, `JSON metadata header followed by randomly-seekable chunks`. 

 - Playcanvas team built upon the SOGv2 format to design their own version of a HLOD hierarchical LOD tiled splats format, named [streamed SOG](https://developer.playcanvas.com/user-manual/gaussian-splatting/formats/streamed-sog/), a binary spatial subdivision where each node geometry is encoded as sog set of webp textures. 

> Spark 2.0 supports SOGS files via a .zip file with manifest.json and referenced .webp images, and also the new extension .sog. However, the new Spark 2.0 file loaders don't support referencing manifest.json via remove URL and fetching the relative URL images. Please download and ZIP the files together and reference that directly for Spark 2.0.

---

It would be great to extend support for tiled splats within sparkjs to not only `RAD`, but also the `streamed SOG` format/specification, which would take advantage of the hierarchical structure of streamed-sog, and the existing sog decoding implementation. 

Note at some point it might as well make sense to also support the `OGC-3dtiles of glTF nodes` packaged with the [KHR_gaussian_splatting](https://github.com/KhronosGroup/glTF/blob/main/extensions/2.0/Khronos/KHR_gaussian_splatting/README.md) and [KHR_gaussian_splatting_compression_spz_2](https://github.com/KhronosGroup/glTF/pull/2531), because an increasing number of implementations can produce such tiled-spalts - within the Cesium/Bentley ecosystem, as well as ESRI or independent tilers).

mrxz:
[...]

PR #389, opened 2026-07-03
Decompress and decode SOG(S) files in JS and add support for unbundled SOG files

mrxz:
This PR builds upon #380 

Instead of decompressing SOG(S) files and decoding their WebP images on the Rust side, this PR moves it to the JS side. This allows the Rust dependencies `zip` and `image` to be omitted from the bundle. Additionally it makes it possible to support unbundled SOG files. For the `build-lod` command the current behaviour is retained as it does not have a JS component. This is handled with a Cargo feature for `spark-lib` that allows to conditionally compile the desired implementation.

The unzipping currently happens synchronously after which the image decoding can start in parallel. An async approach would be nicer, but sadly `fflate` contains a bug preventing this from working reliably (https://github.com/101arrowz/fflate/issues/208).

In terms of bundle size reduction (#321) this PR has the following effect:
| File | Before | After | Delta |
|-----------|----------|--------|-------|
| spark.module.min.js | 2599.85 kB (gzip: 903.24 kB) | 1983.95 kB  (gzip: 644.65 kB) | **-23.7%** (gzip: **-28.6%**) |
| spark.module.js         | 2769.86 kB (gzip: 917.25 kB) | 2163.22 kB  (gzip: 659.79 kB) | **-21.9%** (gzip: **-28.1%**) |        

The unbundled SOG support does require passing the right `SplatFileType` (not auto-detected), but the main reason for reintroducing support is the prospect of supporting Streamed SOG files (#375).

### DIFF
--- a/rust/spark-lib/Cargo.toml
+++ b/rust/spark-lib/Cargo.toml
@@ -7,6 +7,9 @@ license.workspace = true
 authors.workspace = true
 repository.workspace = true
 
+[features]
+sogs_web_decode = []
+
 [dependencies]
 ahash.workspace = true
 anyhow.workspace = true
--- a/rust/spark-lib/src/decoder.rs
+++ b/rust/spark-lib/src/decoder.rs
@@ -8,7 +8,7 @@ use crate::{
     ksplat::KsplatDecoder,
     ply::{PLY_MAGIC, PlyDecoder},
     rad::{RAD_CHUNK_MAGIC, RAD_MAGIC, RadDecoder},
-    sogs::SogsDecoder,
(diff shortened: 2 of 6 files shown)


asundqui:
I love this idea and direction. It solves many things at once: adding back support for unbundled SOG, reduces dependencies and bundle size, and opens the door the streaming SOG support in the future. @oscarlorentzon has been thinking about maybe creating a new subclass of `SplatGenerator` specifically tailored to consume and render SOG-based LoD scans with discrete levels. This will help us move in that direction! Thank you so much @mrxz for this work and sorry for not getting you feedback on this sooner.

Issue #226, opened 2025-11-20
Precise clipping by the camera's near/far planes


soundform:
Spark already trims splats whose centers are outside the camera's frustum. I'd suggest to make a small, but valuable improvement: detect splats that cross the near/far planes and render only the portion of the splat that's inside the frustum. That will involve some math with the error function, which has a simple approximation:

```glsl
  // integrate( exp(-x*x) * 2.0/SQRT_PI, 0..x ) = -1..1
  // https://en.wikipedia.org/wiki/Error_function
  float erf(float x) {
    if (abs(x) > 3.5) return sign(x); // optional
    return sign(x)*sqrt(1. - exp2(-SQRT_PI*x*x));
  }
```

With this change it's possible to compute shadows, add extra light sources and so on. For example, to do a raymarching-style render one can render the scene in thin `(near, far=near+eps)` slices and blend them together:

```js
for (let z = near; z < far; z += eps) {
  camera.near = z;
  camera.far = z + eps;
  await spark.prepare(); // sort splats that are between (z, z+eps)
  await spark.render(); // these layers will be accumulated in the render target
}
```

This will make a difference when too many splats overlap and cause z-order fighting.

Shadows can be computed in a similar manner, but from the perspective of the light source's frustum. For example, to properly illuminate splats with an extra light source, one can set the camera where the new light source is, render 100-1000 slices of the scene, and use the alpha value of the nearest slice as the amount of light reaching the splat:

```js
let shadowMap = new THREE.WebGLArrayRenderTarget(..., ..., numLayers);
let camera = new THREE.PerspectiveCamera(...);
camera.position.set(lightSource.position);

for (let i = 0; i < numLayers; i++) {
  camera.near = i*eps;
  camera.far = (i+1)*eps;
  renderer.setRenderTarget(shadowMap, i);
  await spark.prepare();
  await spark.render(); // only `color.a` component is needed
}

splatMesh.objectModifier = ... // use shadowMap to find the amount of light reaching the splat.xyz
```

Another possibility is to create 3d voxel maps with color densities: use the orthographic camera to render the splats in very thin slices, save volumetric densities as 8 bit rgba, then compress. This might be useful for converting splats to https://code.blender.org/2025/10/volume-grids-in-geometry-nodes.

mrxz:
[...]

PR #397, opened 2026-07-15
fix(renderer): cull splats by projected footprint

yingliang-zhang:
## Summary

- replace center-only lateral `clipXY` rejection with conservative projected-footprint culling for standard 3D Gaussians
- derive the NDC footprint bound from the final ellipse axes after covariance projection, blur, standard-deviation scaling, focal adjustment, and pixel-radius clamping
- keep the 2DGS direct-corner path conservative by relying on GPU primitive clipping instead of rejecting by center
- document the updated `clipXY` semantics and add a focused shader source-contract regression

## Motivation

The vertex shader currently rejects a splat when its center is outside the lateral `clipXY` boundary. A large or anisotropic Gaussian can still overlap the retained region even when its center is outside that boundary, so the center-only test can remove visible border contributions.

This change waits until the projected covariance and final ellipse axes are available. It computes the ellipse's axis-aligned NDC extent and rejects only when the complete footprint misses the configured boundary. A one-projected-pixel padding with a small NDC floor keeps the comparison conservative at float32 boundaries.

## Scope

This is the lateral X/Y part of the broader clipping discussion in #226. It intentionally leaves the existing behind-camera and near/far center tests unchanged; precise Gaussian truncation across near/far planes also interacts with sorting and requires a separate design.

For 2DGS, the existing direct-corner path has no equivalent projected-covariance bound. It now defers lateral clipping to normal GPU primitive clipping rather than retaining the old false-negative center test. The tradeoff is additional vertex work for off-screen 2DGS primitives.

## Verification

- `npm run build:wasm`
- `npm test` — 2/2 passed
- `npm run lint` — 85 files checked, no findings
- `npm run build` — production and development bundles built successfully
- real WebGL browser capture exercised the shader with no compilation or console errors

Related to #226; this PR does not close the near/far clipping request.

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -134,9 +134,9 @@ export interface SparkRendererOptions {
    */
   falloff?: number;
   /**
-   * X/Y clipping boundary factor for Gsplat centers against view frustum.
-   * 1.0 clips any centers that are exactly out of bounds, while 1.4 clips
-   * centers that are 40% beyond the bounds.
+   * X/Y clipping boundary factor, in NDC, for projected Gsplat footprints.
+   * 1.0 retains footprints overlapping the view bounds, while 1.4 retains
+   * footprints overlapping bounds expanded by 40%.
    * @default 1.4
    */
   clipXY?: number;
@@ -648,7 +648,7 @@ export class SparkRenderer extends THREE.Mesh {
       // Modulate Gaussian kernal falloff. 0 means "no falloff, flat shading",
       // 1 is normal e^-x^2 falloff.
       falloff: { value: 1.0 },
-      // Clip Gsplats that are clipXY times beyond the +-1 frustum bounds
+      // Cull projected Gsplat footprints outside clipXY times the +-1 frustum bounds
       clipXY: { value: 1.4 },
       // Debug renderSize scale factor
       focalAdjustment: { value: 1.0 },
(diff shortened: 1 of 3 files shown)


asundqui:
Hi @yingliang-zhang , thank you for your contribution. I agree that the way we are culling splats using just the splat center and `clipXY` isn't ideal, and can lead to large splats "popping" in and out and the boundaries of the screen. Your change does seem to make the culling more conservative and would slightly reduce popping. However, it also comes at the cost of culling fewer splats that aren't visible and slightly more code execution to determine if a splat can be culled.

However, I feel that neither our current solution nor yours really solves the problem on a more fundamental level... The original 3DGS formulation is an approximation with a very strange Z-plane aligned projection that causes problems with clipping and distortion. I think there are several more principled approaches:

- 3DGUT: https://research.nvidia.com/labs/toronto-ai/3DGUT/
- HTGS: https://fhahlbohm.github.io/htgs/
- VKRayGS: https://arxiv.org/html/2503.18682v1

In `SparkRenderer` there is a commented out option `enableRayEval` that we worked on previously (with Florian Hahlbohm of HTGS) that is intended to address the popping+artifacting along the lines of those three approaches. They all share the same per-fragment evaluation of 3DGS, which is more "3D correct".

This is a feature we intend to add to Spark in the near future to improve, stay tuned! cc: @oscarlorentzon  @dmarcos

PR #398, opened 2026-07-15
perf(rust): reduce sort32 bucket overhead

dexhunter:
## Summary

- count valid keys in fixed-size chunks to reduce loop overhead
- reuse one mutable bucket reference per scatter instead of indexing each bucket twice
- return before the scatter passes when there are no active splats
- add regression tests for invalid keys, finite-value boundaries, and stable ordering

## Performance status

The earlier 24.0% latency-reduction headline is withdrawn pending revalidation.

A fresh audit found that the retained evaluator used Node's native WebAssembly runtime but did not record the Node/V8, Rust, wasm-bindgen, or host versions. Its runner contains seven mixed cases, not the ten previously stated, and every case injects invalid keys. It has no all-valid radial-depth fixture or per-layout paired baseline/candidate receipts, so it does not establish an improvement for the common all-valid case.

The required follow-up is a pinned-toolchain, alternating paired rerun with valid-only production-like radial-depth inputs reported separately from mixed-invalid controls, including raw samples and per-layout medians.

Historical autoresearch trace (not validation evidence): https://dashboard.weco.ai/share/4avZ2RktRLkZmJ1sWBU1Tiqqd20Zndg7

## Validation

- cargo test --manifest-path rust/spark-rs/Cargo.toml (2 passed)
- evaluator evidence audit completed; corrected valid-only replay pending
- git diff --check

### DIFF
--- a/rust/spark-rs/src/sort.rs
+++ b/rust/spark-rs/src/sort.rs
@@ -128,21 +128,22 @@ pub fn sort32_internal(
     let Sort32Buffers { readback, ordering, buckets16lo, buckets16hi, scratch } = buffers;
     let keys = &readback[..num_splats];
 
-    // tally low and high buckets (branchless)
+    // tally low and high buckets
     buckets16lo.fill(0);
     buckets16hi.fill(0);
 
     macro_rules! tick {
         ($key:expr) => {{
-            let valid = ($key < DEPTH_INFINITY_F32) as u32;
-            let inv = !$key;
-            let lo = inv & RADIX_MASK;
-            let hi = inv >> RADIX_BITS;
+            if $key < DEPTH_INFINITY_F32 {
+                let inv = !$key;
+                let lo = inv & RADIX_MASK;
+                let hi = inv >> RADIX_BITS;
 
-            // by mask above: lo < 65536 == buckets16lo.len() == RADIX_BASE
-            unsafe { *buckets16lo.get_unchecked_mut(lo as usize) += valid; }
-            // by shift above: hi < 65536 == buckets16hi.len() == RADIX_BASE
-            unsafe { *buckets16hi.get_unchecked_mut(hi as usize) += valid; }
+                // by mask above: lo < 65536 == buckets16lo.len() == RADIX_BASE
+                unsafe { *buckets16lo.get_unchecked_mut(lo as usize) += 1; }
+                // by shift above: hi < 65536 == buckets16hi.len() == RADIX_BASE
+                unsafe { *buckets16hi.get_unchecked_mut(hi as usize) += 1; }
+            }
         }};
     }
 
@@ -167,6 +168,10 @@ pub fn sort32_internal(
     let active_splats = prefix_sum_exclusive(buckets16lo);
     prefix_sum_exclusive(buckets16hi);
 
+    if active_splats == 0 {
+        return Ok(0);
+    }
+
     // ——— Pass #1: bucket by inv(low 16 bits) ———
 
     // scatter into scratch by low bits of inv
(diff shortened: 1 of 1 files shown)


dexhunter:
Rechecked against current `main` (`fec6d05`): this still merges cleanly, and `rust/spark-rs/src/sort.rs` has not changed upstream since #327, so the patch is still confined to that one file (+53/-16).

The amber check state is not a failure — `Spark CI Linux` and `Spark CI Windows` are both sitting in `action_required` awaiting fork-workflow approval, so they have never actually run. I ran the same steps locally against `main` plus this patch: `npm run lint` clean across 84 files, `npm run test` 1/1, and `cargo test --manifest-path rust/spark-rs/Cargo.toml` 2/2 (the two regression tests this PR adds).

Separately, and this applies to any Rust-only change rather than to this PR specifically: `build-dist.yml` triggers on `paths: ['src/**']`, so a commit confined to `rust/` does not regenerate the committed `dist/` bundles that inline the wasm. #327 landed on 2026-06-08 and the next `Bump dist build` was 2026-07-15 (`8a033bc1`), immediately after the unrelated `src/` change in #378 — so that sort work reached `dist/` about five weeks after it merged. Happy to add `rust/**` to that trigger, here or as a separate PR, if that would be useful — or to leave it alone if the dist refresh is meant to be deliberate.

mrxz:
Apologies for the late response, looking into the linked "Autoresearch trajectory" I'm not so sure about the stated 24.0% latency reduction. Would it be possible to know the actual test data/suite and which runtime/engine was used to execute these test, as WASM performance can differ quite a bit between browsers/runtimes.

The largest gain seems to come from the conditional when tallying. In practice there really shouldn't be many (if any) invalid depth values, as the default sort metric is based on radial depth (always positive). Seeing it have such an impact is somewhat unexpected. If it doesn't have any negative performance impact on the common case, it might still be worth considering, of course.

> Separately, and this applies to any Rust-only change rather than to this PR specifically: `build-dist.yml` triggers on `paths: ['src/**']`, so a commit confined to `rust/` does not regenerate the committed `dist/` bundles that inline the wasm.

Good point, created a PR #412

dexhunter:
*This is Aiden on behalf of Dex*

You're right to question the 24.0% figure. I audited the retained evaluator instead of inferring from its aggregate result. It used Node's native WebAssembly runtime, but the artifacts do not record the Node/V8, Rust, wasm-bindgen, or host versions. More importantly, the retained runner contains seven cases—not the ten stated in the PR—and every case injects invalid keys at fixed intervals. It has no all-valid radial-depth fixture or per-layout paired baseline/candidate receipts, so the current evidence does not support that headline for the common all-valid case.

I'm treating the 24.0% claim as unverified. The required follow-up is a pinned-toolchain, alternating paired rerun with valid-only production-like radial-depth inputs, reported separately from mixed-invalid controls, with raw samples and per-layout medians. I'll update the performance claim from that evidence before asking you to evaluate the optimization.

asundqui:
Hi @dexhunter , thank you for your contribution and trying to improve the sort time, which of course is one of the most important aspects of a 3DGS renderer! From the above context it sounds like the improved performance was done using an old build, without the improvements in #327, and I suspect this PR doesn't improve things in a meaningful way.

We welcome these sorts of improvements, but perhaps we should close this PR for now? @mrxz @dmarcos @oscarlorentzon if any of you agrees, let's close this PR?

dexhunter:
Thanks — I checked this carefully. One correction: the PR base (750812d) is three commits ahead of the merge of #327, and rust/spark-rs/src/sort.rs is unchanged from that base through current main, so the original run was not against pre-#327 source. However, the retained benchmark has only seven mixed-invalid fixtures and lacks valid-only radial-depth cases, paired receipts, and pinned runtime/toolchain provenance. I cannot substantiate the 24% claim for the common case, and I do not want to ask maintainers to review an unproven optimization. I am closing this for now. If I revisit it, I will submit a ready-for-review PR backed by current-main paired measurements with valid and invalid workloads reported separately.

PR #403, opened 2026-07-29
Support scaled cameras (WebXR rig scaling)

marwie:
Splat scenes break under a uniformly **scaled camera**. In [Needle Engine](https://needle.tools) we scale the user's XR **rig** — not the scene — when users resize a placed scene in AR/VR: world coordinates stay stable, which keeps networked multi-user sessions consistent (every participant shares the same scene space while viewing at their own scale). Under such a scaled camera, three issues stack up in Spark; all fixed here, and **with an unscaled camera every change is a no-op** (scale factor 1 throughout).

1. **Render transform drops the camera scale.** 
The accum→view transform is decomposed to pos + quat; the translation keeps the scale factor but the rotated centers and splat extents don't → the scene warps instead of scaling. Fix: `renderToViewScale` uniform from the previously-discarded decompose scale (`scales *= s`, `viewCenter = s * quatVec(q, center) + pos`). The `enableCovSplats` branch already uses the full basis and is unchanged; `OldSparkRenderer` untouched.

2. **Sort/LOD triggers are world-unit constants.** 
The view-changed epsilon (1 mm) and LOD ramp (1 m) don't account for camera scale: at rig scale *s*, physical head movement maps to *s×* less world movement — a scene scaled up 20× never re-sorts (frozen blend order, foveation holes). Fix: both thresholds scale with the camera's world scale. The LOD pixel cut is deliberately **not** scaled — perspective cancels uniform camera scale for size-at-distance (device-verified: scaling it double-counts).

3. **XR updates read the ArrayCamera container outside the frame.** `
updateInternal` runs on `renderer.xr.getCamera()`, often in a `setTimeout`: the parentless container's `getWorldPosition()` recomputes `matrixWorld` from the raw reference-space local matrix, discarding the rig transform — rendering is correct, but `sortedCenter`/`viewToWorld` live in raw head space. Fix: use the per-eye camera (`getCamera().cameras[0] ?? getCamera()`), which engines can keep validly composed at all times.

Found and device-verified building AR splat viewing on Needle Engine (Pixel 9, Chrome `immersive-ar`): pinch-scaling a placed splat scene now scales correctly and keeps sorting/LOD alive at any rig scale.

Demo scene: https://splat-z23hmxbhac3d-zjprnz.needle.run/ (Enter AR and scale the scene with two fingers)

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -22,6 +22,14 @@ import {
   uploadU32DataTextureRows,
 } from "./utils";
 
+const renderToViewScaleTmp = new THREE.Vector3();
+
+// Average (uniform) world scale of a camera
+function getCameraWorldScale(camera: THREE.Camera): number {
+  const s = camera.getWorldScale(renderToViewScaleTmp);
+  return (s.x + s.y + s.z) / 3;
+}
+
 export interface SparkRendererOptions {
   /**
    * Pass in your THREE.WebGLRenderer instance so Spark can perform work
@@ -621,6 +629,8 @@ export class SparkRenderer extends THREE.Mesh {
       renderToViewQuat: { value: new THREE.Quaternion() },
       // SplatAccumulator to view transformation translation
       renderToViewPos: { value: new THREE.Vector3() },
+      // SplatAccumulator to view transformation uniform scale (camera world scale)
+      renderToViewScale: { value: 1 },
       renderToViewBasis: { value: new THREE.Matrix3() },
       renderToViewOffset: { value: new THREE.Vector3() },
       // Maximum distance (in stddevs) from Gsplat center to render
@@ -780,8 +790,13 @@ export class SparkRenderer extends THREE.Mesh {
     accumToCamera.decompose(
       this.uniforms.renderToViewPos.value,
       this.uniforms.renderToViewQuat.value,
-      new THREE.Vector3(),
+      renderToViewScaleTmp,
     );
+    this.uniforms.renderToViewScale.value =
+      (renderToViewScaleTmp.x +
+        renderToViewScaleTmp.y +
+        renderToViewScaleTmp.z) /
+      3;
     this.uniforms.renderToViewBasis.value.setFromMatrix4(accumToCamera);
 
     this.uniforms.maxStdDev.value = spark.maxStdDev;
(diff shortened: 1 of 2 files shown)


asundqui:
Hi @marwie, thank you for this PR. I had never considered this use case, and it makes sense what you want to do.

However, it looks like in Three.js r183 they removed the ability to scale the camera: https://github.com/mrdoob/three.js/pull/32805

Although your PR looked pretty clean and works, I think in light of what Three.js is doing we should probably not support scaling the camera either - it would not work with anything else rendered in the scene.

As an alternative, you could put all your scene objects into a `Group` and simply scale that inversely. There may be some translation also involved to achieve the effect you want, but it should be equally expressive as scaling the camera.

Do you agree @oscarlorentzon @dmarcos ? If so should we close this PR?

asundqui:
Closing because we don't think we should support scaled cameras as per above.

marwie:
Hello @asundqui I believe there's a misunderstanding. 

Three.js didn't remove the ability to scale the camera or have a scaled camera in a scene but they changed how scaling affects rendering. Previously three's camera scale modified view transformation (which caused side-effects like wrong lighting intensity) - **camera or rig still may have a world scale**.

mrxz:
As @marwie says, it is indeed still supported in Three.js to have an effective (uniform) scale on the camera. 

> As an alternative, you could put all your scene objects into a `Group` and simply scale that inversely. There may be some translation also involved to achieve the effect you want, but it should be equally expressive as scaling the camera.

This actually won't be identical, for example when it comes to physical lighting. A crude example, but if you scale a scene with several lights, the lights would effectively be moved further apart _without_ their physical properties (e.g. intensity) being scaled up to match. However, if you'd apply the inverse scale to the camera rig the scale on the view matrix gets excluded, keeping the lighting calculations consistent.

The question still remains if Spark should/will support this, but it is supported in Three.js and for WebXR it is the most convenient way to scale the spectator when it comes to things like physical lights.

marwie:
Another reason why we can't simple reparent is that our whole networking stack is built around multiuser experiences where some users open the same scene in AR, some in VR and some on desktop in one shared environment. Here different devices and views have different camera rigs and perspectives on the same scene.

For example AR users - who often have a birds-eye view on the scene when they place the 3D environment on a small table - are parented in a scaled XR rig. Exactly this one rig position and scale is networked and used for synchronizing views. Nothing else in the scene is affected or needs to be changed for this to work.

I hope we can merge this PR (or fix it in another way) in sparkjs :)

dmarcos:
reopening for further discussion

Issue #408, opened 2026-08-05
SparkRenderer.dispose() does not dispose its geometry — 2 WebGL buffers leak per instance (v2.1.0)


tyirenkyi:
## Summary

`SparkRenderer.dispose()` does not dispose the geometry of its own `THREE.Mesh`, so **2 WebGL buffers leak per `SparkRenderer` instance**. Applications that mount and unmount splat scenes repeatedly (route changes, toggling a render mode, opening/closing an editor) accumulate GPU buffers for the lifetime of the WebGL context.

Confirmed by a one-line change on the caller side: adding `spark.geometry.dispose()` after `spark.dispose()` takes the growth to exactly zero.

## Environment

- `@sparkjsdev/spark` **2.1.0**
- three.js **r180**
- Chrome (stable, real GPU via `--use-gl=angle`), macOS

## Reproduction

Mount a `SparkRenderer` + `SplatMesh`, render one frame, tear both down with the public API, repeat. Live WebGL objects are counted by wrapping `createBuffer`/`deleteBuffer` on `WebGL2RenderingContext.prototype` before anything else runs, so an object created and never deleted stays counted.

```js
for (let i = 0; i < CYCLES; i++) {
  const spark = new SparkRenderer({ renderer });
  scene.add(spark);

  const mesh = new SplatMesh({ fileBytes: bytes.slice(0), fileType });
  await mesh.initialized;
  scene.add(mesh);

  renderer.render(scene, camera);          // make sure everything is allocated + uploaded
  await new Promise(requestAnimationFrame);

  scene.remove(mesh);
  mesh.dispose();
  scene.remove(spark);
  spark.dispose();                         // <-- everything the public API offers
  // spark.geometry.dispose();             // <-- uncomment and the leak disappears
}
```

## Result

Live WebGL buffers, sampled after a forced GC each cycle (8 cycles, 500k-Gaussian `.ply`):

| teardown | live buffers per cycle | per cycle |
| --- | --- | --- |
| `spark.dispose()` | 4 → 6 → 8 → 10 → 12 → 14 → 16 → 18 | **+2** |
| `spark.dispose()` + `spark.geometry.dispose()` | 2 → 2 → 2 → 2 → 2 → 2 → 2 → 2 | **0** |

Live textures are flat at 7 in both, so this is specific to the geometry's buffers.

## Where they are allocated

[...]

PR #410, opened 2026-08-16
Dispose geometry and material in SparkRenderer.dispose()

ArisLiWind:
## Summary

Fixes a GPU resource leak in `SparkRenderer.dispose()` by releasing the renderer-owned `SplatGeometry` created in the constructor.

## Root cause

`SparkRenderer` constructs its own `SplatGeometry`:

```ts
const geometry = new SplatGeometry();
super(geometry, material);
```

That geometry is owned by the renderer instance, but `dispose()` previously released render targets, textures, workers, accumulators, and pager state without disposing the geometry itself.

Because `SplatGeometry` allocates both a `position` attribute and an `index` buffer, each create/dispose lifecycle could leave two WebGL buffers behind.

## Fix

Store the constructor-created geometry as renderer-owned state and dispose it during `SparkRenderer.dispose()`.

## Regression coverage

Added tests to verify that:

- `SparkRenderer.dispose()` releases its internal geometry
- disposing a renderer does not dispose geometry assigned from outside
- `dispose()` is safe to call multiple times

## Validation

- `npm test`
- `npm run lint -- --reporter=summary`


mrxz:
> Store the constructor-created geometry as renderer-owned state and dispose it during `SparkRenderer.dispose()`.

There is no need to specifically keep track of the constructor-created geometry. Replacing the geometry of the `SparkRenderer` is not supported. You can simplify the dispose logic to just `this.geometry.dispose()`. For completion sake could you also dispose the material?

Also no need to include the unit tests in this PR. While ultimately we should have extensive tests covering things like this, the required bootstrapping isn't ideal.

asundqui:
@mrxz this will already be covered by #413 right? If so should we close this PR?

oscarlorentzon:
This is independent from #413 which disposes the `Object3D`, and as @mrxz mentioned we should remove the tests for now and simplify this to:

```ts
this.geometry.dispose();
this.material.dispose();
```

asundqui:
I see, got it! @ArisLiWind how would you like to make these changes so we can merge the PR? We can also take over this PR and apply these finishing touches.

ArisLiWind:
Thanks! Feel free to take over the PR and apply the remaining changes needed for the merge.@asundqui

oscarlorentzon:
Took this over as discussed, thanks @ArisLiWind.

Simplified according to the discussion above and rebased on main to resolve the conflict with #413.

oscarlorentzon:
Fixes #408

oscarlorentzon:
@dmarcos This should be ready to merge.

dmarcos:
thanks!

PR #411, opened 2026-08-22
Add reflected SH support for MirrorGaussian scenes

katjaschwarz:
## Summary
- Add an object-space view basis to `SplatMesh` for reflected spherical-harmonic evaluation.
- Apply the basis consistently to packed, extended, and paged/LOD splat sources.
- Add a manifest-driven multi-mirror example that composites real, reflected, and mask passes from one camera snapshot.

## Why
MirrorGaussian reflects Gaussian geometry across each mirror plane, but view-dependent color also needs the corresponding reflection applied to the SH view direction. Without this transform, reflected geometry has incorrect view-dependent appearance.

## How
`SplatMesh` exposes a `shViewBasis` matrix and passes it into every directional splat fetch path before SH evaluation. The example loads one reflected scene and mask per manifest mirror, waits for each Spark renderer to sort against the same camera snapshot, and composites the resulting render targets. It disables LOD in the example because independently updated LOD trees can put the passes at different camera states.

## Test plan
- [x] `npm run lint`
- [x] `npm test`
- [x] `npm run build`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/examples/mirrorgaussian/README.md
+++ b/examples/mirrorgaussian/README.md
@@ -1,0 +1,43 @@
+# MirrorGaussian
+
+This example composites a real-scene splat pass with one reflected splat and
+mask pass per mirror. It expects SPZ files beside a `mirrorgaussian-spark-v2`
+manifest:
+
+```text
+assets/
+  mirror.json
+  real.spz
+  mirrors/
+    <mirror-id>/
+      mask.spz
+      mirror.spz
+```
+
+The manifest's asset paths may use `.ply`; the example substitutes `.spz` when
+loading them. Each mirror entry must provide `sh_view_basis_row_major`, a 3x3
+object-space basis used to evaluate the reflected scene's spherical harmonics.
+
+```json
+{
+  "schema": "mirrorgaussian-spark-v2",
+  "assets": { "real": "real.ply" },
+  "mirrors": [
+    {
+      "id": "1",
+      "assets": {
+        "mask": "mirrors/1/mask.ply",
+        "mirror": "mirrors/1/mirror.ply"
+      },
+      "sh_view_basis_row_major": [-1, 0, 0, 0, 1, 0, 0, 0, 1]
+    }
+  ]
+}
+```
+
+The passes use one camera snapshot and finish sorting before their render
+targets are composited. LOD is disabled because independently updated LOD trees
+can make the real, reflected, and mask passes use different camera states.
+
+Run `npm start`, then open
+`http://localhost:8080/examples/mirrorgaussian/`.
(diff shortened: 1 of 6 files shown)


mrxz:
The descriptions reference a `MirrorGaussian` which isn't part of the PR and the example refers to assets that aren't included. It would be nice to have a small and self-contained example showing a mirrored splat for which the spherical harmonics need to be reflected (e.g. splat with a mirrored version next to it)

Since you appear to be loading dedicated mirror and mask splat files anyway, wouldn't it make more sense to mirror the SH ahead of time? And even if mirrored at runtime, wouldn't it be an option to mirror the SH before evaluating?

Not saying I'm against the idea of handling it at the SH evaluation stage, but then I'd rather make the SH evaluation pluggable/extendable.

asundqui:
@katjaschwarz thank you so much for this PR. It's an interesting use case, and I understand why you would want to do this. If the mirror transform is "simple", for example reflection across X/Y/Z-axis, I believe you could mirror the SH coefficients ahead of time, but otherwise it seems that rotating/mirroring the SH coefficients would "lose information" because we'd essentially be resampling them (@mrxz your question in the previous comment). So being able to transform the SH basis in real-time would be required to do this "correctly".

That being said, there are a few things that lead me to conclude we probably shouldn't merge this in:
1) It's a very niche use case
2) It makes some of the basic internal interfaces a bit more complicated for such niche use
3) I think this is actually already doable in Spark using the "CovSplats" (short for covariance splats) encoding variant.

To explain a bit better: Spark originally had PackedSplats, and then we added ExtSplats, and for both of those we explicitly store the splat's major x/y/z scales and orientation. This representation is nice and is used everywhere in the pipelines (for example `SplatMesh.objectModifiers/worldModifiers`), and expect to be able to operate on scales+orientations along with other parameters. Unfortunately, this means that applying a non-uniform scale to a entire SplatMesh is difficult, since each splat is independently oriented and we can't just multiply the scales. You can get around this by computing the 3D covariance, applying the transformation, and then diagonalizing the resulting 3Dcov, but this is an expensive operation.

We added `CovSplats` as a side option for people who wanted non-uniform scaling across objects. It's a different pipeline than the normal flow so users have to pick one or the other via `SparkRenderer.covSplats`. I believe CovSplats can also handle mirror transformations, and also that this transformation would be correctly applied when evaluating SH coefficients, because it looks like all the transforms use the full transform matrix rather than decomposed rotation/translation/uniform scale like the regular splat pipelines do.

So with all this said, I think we should probably close this PR, I hope that makes sense! @mrxz @oscarlorentzon @dmarcos do you agree?

mrxz:
Since it's mirrored through a plane, it should always be possible to do the "simple" mirror transform and then position/rotate it into place. In any case, I agree with @asundqui that this PR should be closed, as there are other ways to achieve this without these changes.

PR #413, opened 2026-08-26
Call `super.dispose()` in Object3D subclasses for Three.js r186

mrxz:
There are a few descendant classes of Object3D in Spark that have their own `dispose` method. In the upcoming r186 release of Three.js, the base class will have a `dispose` method as well, so derived classes should call it (see https://github.com/mrdoob/three.js/pull/34141).

This PR adds conditional calls to the base `dispose` method if present. This way the same code will work for the currently supported Three.js versions as well as the upcoming r186 and later. Since the types don't know about this new `dispose` method the `@ts-ignore` annotations are required.

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -672,6 +672,9 @@ export class SparkRenderer extends THREE.Mesh {
   }
 
   dispose() {
+    // @ts-ignore Object3D has a dispose method in Three.js >= r186
+    super.dispose?.();
+
     if (this.target) {
       this.target.dispose();
       this.target = undefined;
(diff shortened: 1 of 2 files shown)


oscarlorentzon:
Good to have early, a no-op until r186. One thing:

- [src/SparkRenderer.ts:675](https://github.com/sparkjsdev/spark/blob/201dafb0f0e740ed4eebc8aafba7ccc9715b209f/src/SparkRenderer.ts#L675) and [src/SplatMesh.ts:566](https://github.com/sparkjsdev/spark/blob/201dafb0f0e740ed4eebc8aafba7ccc9715b209f/src/SplatMesh.ts#L566), the comment says "Object base class", it should be "Object3D".

Could you fix that?

mrxz:
Updated the comments to explicitly mention `Object3D`.

oscarlorentzon:
Looks good. @dmarcos do you want to merge?

PR #414, opened 2026-08-27
Add `lint:rust` command and resolve various clippy linting warnings

mrxz:
Currently only the TypeScript code is automatically linted and formatted, whereas the Rust code is left as-is. Ideally we'd apply the same automatic process to both. This helps avoid common mistakes/pitfalls and ensures consistent code formatting. Related https://github.com/sparkjsdev/spark/pull/370

This PR introduces a `lint:rust` script to easily execute clippy. This is _not_ included in the pre-commit hook nor in any of the CI pipelines yet. In order for that to be useful, all pre-existing linting warnings should be resolved (either fixed or explicitly allowed).

Additionally various linter warnings have been resolved. Though the following rules have been skipped as they (subjectively) reduce the readability or are non-trivial:
| Rule name | Notes |
|---|---|
| [clippy::identity_op](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#identity_op) | Many instances where a consecutive set of elements is computed/referenced, in which case the `+ 0` is nice for symmetry. |
| [clippy::len_without_is_empty](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#len_without_is_empty) | |
| [clippy::new_without_default](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#new_without_default) | |
| [clippy::needless_range_loop](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#needless_range_loop) | |
| [clippy::excessive_precision](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#excessive_precision) | Some of the SH constants trigger this warning, but since these literally match the JS and GLSL code, leaving them unchanged is preferred |
| [clippy::too_many_arguments](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#too_many_arguments) | Part of the exported functions of the WASM blob, so changing this would require further changes |
| [clippy::let_and_return](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#let_and_return) | See https://github.com/sparkjsdev/spark/pull/370#issuecomment-4690217960 |
| [clippy::clone_on_copy](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#clone_on_copy) | |
| [clippy::type_complexity](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#type_complexity) | |
| [clippy::wrong_self_convention](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#wrong_self_convention) | |
| [clippy::collapsible_if](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#collapsible_if) | |
| [clippy::needless_return](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#needless_return) | In case of early returns in the same method body, I'd have a slight preference in consistently using `return` |
| [clippy::some_filter](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#some_filter) | |
| [clippy::large_enum_variant](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#large_enum_variant) | |
| [clippy::only_used_in_recursion](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#only_used_in_recursion) | |
| [clippy::manual_range_contain](https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#manual_range_contains) | Frankly the explicit `version < 1 || version > 3` reads a lot nicer than `!(1..=3).contains(&version)` |

The `needless_range_loop` is interesting in the regard that it can help avoid bound checks. Though quite often the index is still needed for offset in the source or destination vectors. The resulting code would not always be as easy to read, especially when the iterated collection is the destination and not the source.

In any case, I do think this is a good start. Follow up steps would be to settle on which linting rules to include and which to exclude as well as integrating it in the CI pipelines + commit hook (preferably only if/when rust code has changed given how slow these Rust tools are).

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "lint": "biome check .",
     "lint:fix": "biome check . --fix",
     "lint:summary": "biome check --reporter=summary",
+    "lint:rust": "cargo clippy --manifest-path rust/Cargo.toml",
     "site:build:relativeurls": "npm run build && npm run site:clean && node scripts/copy-site-files.js && mkdocs build && node scripts/rename-assets-to-static.js && node scripts/replace-urls.js",
     "site:build": "npm run build && npm run site:clean && node scripts/copy-site-files.js && mkdocs build && node scripts/rename-assets-to-static.js",
     "site:clean": "node scripts/clean-site-files.js && node scripts/clean-site.js",
--- a/rust/build-lod/src/main.rs
+++ b/rust/build-lod/src/main.rs
@@ -97,8 +97,8 @@ fn process_file_lod(filename: &str, options: &BuildLodOptions) {
 }
 
 fn process_file_lod_tsplat<TS: SplatReceiver + TsplatArray + SplatGetter>(filename: &str, options: &BuildLodOptions, splats: TS) {
-    let mut decoder = MultiDecoder::new(splats, None, Some(&filename));
-    let mut splats = match read_file_chunks(&filename, &mut decoder) {
+    let mut decoder = MultiDecoder::new(splats, None, Some(filename));
+    let mut splats = match read_file_chunks(filename, &mut decoder) {
         Ok(_) => {
             println!("Detected file type: {:?}", decoder.file_type.unwrap());
             decoder.into_splats()
--- a/rust/spark-lib/src/bhatt_lod.rs
+++ b/rust/spark-lib/src/bhatt_lod.rs
@@ -19,7 +19,7 @@ pub fn compute_lod_tree<TA: TsplatArray>(splats: &mut TA, lod_base: f32, logger:
 
     splats.sort_by(|s| s.feature_size());
     splats.prepare_children();
-    logger(&format!("Sorted and prepared splats"));
+    logger("Sorted and prepared splats");
 
     let mut is_active = Vec::with_capacity(splats.len() * 2 - 1);
     is_active.resize(splats.len(), true);
@@ -235,7 +235,7 @@ pub fn compute_lod_tree<TA: TsplatArray>(splats: &mut TA, lod_base: f32, logger:
         if next_frontier.is_empty() {
             break;
         }
-        limit_size = limit_size / 4.0;
+        limit_size /= 4.0;
         frontier = next_frontier;
     }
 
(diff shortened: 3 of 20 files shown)


asundqui:
@mrxz As much as I love writing and formatting code according to my aesthetics, I chatted with @oscarlorentzon and in that past chatted with @dmarcos about this. I think most people prefer automatic formatting and requiring lint pass before merging. I am okay with. As a first step cleaning up the lints gets us part there. In the near future perhaps we do a rustfmt pass as well and get all the formatting changes in in one PR? WDYT @dmarcos should we move to a pre-commit hook for Rust code as well?

oscarlorentzon:
As @asundqui mentioned this is a good direction. A few things regarding these changes specifically:

- [package.json:35](https://github.com/sparkjsdev/spark/blob/ca76e6af35f7e6833bc5c2218f21725371df6b87/package.json#L35): without `--all-targets` the script skips `#[cfg(test)]`, so its report is incomplete.
  - [rust/spark-lib/src/ordering.rs:153](https://github.com/sparkjsdev/spark/blob/ca76e6af35f7e6833bc5c2218f21725371df6b87/rust/spark-lib/src/ordering.rs#L153) still has `(shift % 64) as u32`, the same cast already fixed at [:62](https://github.com/sparkjsdev/spark/blob/ca76e6af35f7e6833bc5c2218f21725371df6b87/rust/spark-lib/src/ordering.rs#L62), and `unnecessary_cast` is not in the skip table.
- Three `copy_from_slice` calls are missing a trailing semicolon: [rust/spark-rs/src/ext_splats.rs:269](https://github.com/sparkjsdev/spark/blob/ca76e6af35f7e6833bc5c2218f21725371df6b87/rust/spark-rs/src/ext_splats.rs#L269), [:765](https://github.com/sparkjsdev/spark/blob/ca76e6af35f7e6833bc5c2218f21725371df6b87/rust/spark-rs/src/ext_splats.rs#L765) and [:816](https://github.com/sparkjsdev/spark/blob/ca76e6af35f7e6833bc5c2218f21725371df6b87/rust/spark-rs/src/ext_splats.rs#L816).
- `needless_return` is on the skip list for bodies with early returns, but it was applied at [rust/spark-lib/src/rad.rs:1799](https://github.com/sparkjsdev/spark/blob/ca76e6af35f7e6833bc5c2218f21725371df6b87/rust/spark-lib/src/rad.rs#L1799), and that body has an early return at [:1791](https://github.com/sparkjsdev/spark/blob/ca76e6af35f7e6833bc5c2218f21725371df6b87/rust/spark-lib/src/rad.rs#L1791).

@mrxz Shoule we fix those before merging?

PR #415, opened 2026-08-27
Make members of `SplatEncoding` non-optional to avoid repeated fallback handling

mrxz:
The `SplatEncoding` type was used both as the actual encoding values as well as the type for specifying non-default encoding values. This meant that each member had to be optional, which in turn meant a lot of code repeated the same default fallback values as the type suggested it could be undefined.

In practice however, these values would always be set and these fallbacks were just repeated to appease the type system. This PR changes the typing of `SplatEncoding` such that all members are required, removing the need for implementing fallbacks throughout the code base. For the `PackedSplats` and `SplatMesh` constructors a `Partial<SplatEncoding>` is used so these can still receive only the encoding properties that need to be overridden.

Additionally this PR contains some tightening of a few "result" types, marking their members as readonly for good measure, as well as the removal of two unused constants `USE_COMPILED_PARSER_FUNCTION` (obsolete with #374) and `WASM_SPLAT_SORT` (obsolete with #322)


asundqui:
@mrxz  this is a long overdue cleanup, I agree this is much better than all the optionals and fallbacks we had in that past. Thank you for your attention to it! @oscarlorentzon @dmarcos this is a good idea right?

oscarlorentzon:
We might want to handle the remaining default assignments at [`PackedSplats.ts:1066`](https://github.com/sparkjsdev/spark/blob/d00fe62545dbee89545c046d52c0a07b70c4c962/src/PackedSplats.ts#L1066)  and [`SplatMesh.ts:1088`](https://github.com/sparkjsdev/spark/blob/d00fe62545dbee89545c046d52c0a07b70c4c962/src/SplatMesh.ts#L1088). We can fallback with`?? DEFAULT_SPLAT_ENCODING` and then read the members off the object.

Should the `(default: undefined)` comments at [`PackedSplats.ts:88`](https://github.com/sparkjsdev/spark/blob/d00fe62545dbee89545c046d52c0a07b70c4c962/src/PackedSplats.ts#L88) and [`SplatMesh.ts:116`](https://github.com/sparkjsdev/spark/blob/d00fe62545dbee89545c046d52c0a07b70c4c962/src/SplatMesh.ts#L116) name `DEFAULT_SPLAT_ENCODING` now?

mrxz:
> We might want to handle the remaining default assignments at [`PackedSplats.ts:1066`](https://github.com/sparkjsdev/spark/blob/d00fe62545dbee89545c046d52c0a07b70c4c962/src/PackedSplats.ts#L1066) and [`SplatMesh.ts:1088`](https://github.com/sparkjsdev/spark/blob/d00fe62545dbee89545c046d52c0a07b70c4c962/src/SplatMesh.ts#L1088). We can fallback with`?? DEFAULT_SPLAT_ENCODING` and then read the members off the object.

Addressed both of these as well, does nicely get rid of some more repeated default values. Also eliminated them from the initial values of the uniforms, as the `update` method already takes care of setting it before rendering.

> Should the `(default: undefined)` comments at [`PackedSplats.ts:88`](https://github.com/sparkjsdev/spark/blob/d00fe62545dbee89545c046d52c0a07b70c4c962/src/PackedSplats.ts#L88) and [`SplatMesh.ts:116`](https://github.com/sparkjsdev/spark/blob/d00fe62545dbee89545c046d52c0a07b70c4c962/src/SplatMesh.ts#L116) name `DEFAULT_SPLAT_ENCODING` now?

From the user's POV nothing should have changed. While it's slightly confusing that it states `undefined`, I'm not sure listing `DEFAULT_SPLAT_ENCODING` is going to make it clearer at this point. For one there's no reference to it in the docs currently and secondly the comment indicates it's to _"override the default splat encoding ranges"_. In other words it would then imply that the default value for the _override_ of the default splat encoding ranges is the default splat encoding values.

I'd rather leave the docs improvements for future PRs. Ideally we'd move to JSDoc based comments throughout, ensuring people get these comments straight in their IDE, and in time hopefully automatically generate the relevant (reference) docs from it.

oscarlorentzon:
Leaving the docs as is for now makes sense. Looks good to merge @dmarcos.

dmarcos:
thanks

PR #417, opened 2026-08-31
Tree shaking improvements

mrxz:
This PR improves the tree-shakability of Spark by focussing on classes that users might not need or use. See #321 

* Ensures the `SplatSkinning` class is only included when used, including its shader chunks.
* Ensures `SplatEdit` and `SplatEditSdf` are tree-shaken, unless imported.
  * This is done by adopting the three.js approach of having `isType = true` properties on the classes for type discrimination. Since these can't easily be used in TypeScript, however, helper type guard functions are introduced (`isSplatEdit` and `isSplatEditSdf`).
  * NOTE: The `SplatEdits` class is still included at all times as `SplatMesh` is capable of constructing this: [`SplatMesh.ts#975`](https://github.com/sparkjsdev/spark/blob/1be70d14e81db120f266371e6c84cdd46f158291/src/SplatMesh.ts#L975-L978)
* Ensures `SparkXr` and related code is tree-shaken.
  * There was a lot of duplication between `SparkXr.ts` and `hands.ts`, but now the former imports from the latter
  * Annoyingly a simple `.length` prevented tree-shaking, so resorted to a reduce function which can be statically analyzed for elimination (alternatively an IIFE could be used for this as well)


src/hands.ts
```diff
@@ -75,9 +81,9 @@ export const JOINT_INDEX: { [key in JointId]: number } = {
 
 export const JOINT_RADIUS: { [key in JointId]: number } = {
   w: 0.02,
-  t0: 0.02,
-  t1: 0.014,
-  t2: 0.0115,
+  t0: 0.015,
+  t1: 0.012,
+  t2: 0.0105,
```
asundqui, line 86: Curious about these changes? Do they match another part of the code or a previous source? I forget but I think I may have edited them from the original data I used. Don't feel strongly either way but just curious!

asundqui:
This is great, anything that reduces bundle size and opts-in code sounds good to me! @oscarlorentzon @dmarcos we should merge this yes?

src/hands.ts
```diff
@@ -14,36 +14,42 @@ const DEFAULT_MOVE_INERTIA = 0.5;
...
-  r2 = "ring-finger-phalanx-intermediate",
-  r3 = "ring-finger-phalanx-distal",
-  r4 = "ring-finger-tip",
-  p0 = "pinky-finger-metacarpal",
-  p1 = "pinky-finger-phalanx-proximal",
-  p2 = "pinky-finger-phalanx-intermediate",
-  p3 = "pinky-finger-phalanx-distal",
-  p4 = "pinky-finger-tip",
-}
+export const JointEnum = {
```
oscarlorentzon, line 17: JointEnum is exported from index.ts, so this seems to break compatibility for anyone currently using `let j: JointEnum`. Do we need handle that for minor version releases?

oscarlorentzon:
> This is great, anything that reduces bundle size and opts-in code sounds good to me! @oscarlorentzon @dmarcos we should merge this yes?

Looks good. It would be good to clarify the potential API compatibility issue before we merge.

src/hands.ts
mrxz: There were two almost identical lists, one in `hands.ts` and one in `SparkXr.ts`, the latter seemed more recent so those values are now used. They seemed close enough that either one would probably be fine.

One alternative could be to retrieve the joint radii from the WebXR API (https://www.w3.org/TR/webxr-hand-input-1/#dom-xrframe-filljointradii), as that could in theory align better with the user's actual hands (bar spec mandated anonymization)

src/hands.ts
mrxz: Frankly I doubt people use the `JointEnum` type, as it's effectively just the WebXR Hand Input Module joint names. The interaction with the WebXR API is handled by the `SparkXr`/`hands.ts` code and from the user's code everything is keyed on `JointId`.

Mentioning the change in the release notes should suffice IMHO.

oscarlorentzon: Agree, a mention should be enough.

dmarcos:
good to go?

dmarcos:
thanks!

PR #419, opened 2026-09-02
Add gsplatXaxis/Yaxis/Zaxis helper dynos

asundqui:
The dyno helper `gsplatNormal` computes the minimum of the X/Y/Z splat axis and returns a unit vector in that direction, which in some cases can be used as a proxy for a "normal". However, sometimes your splats are always oriented such that the Z-axis points in that direction. In that case it can be useful to have an additional dyno helper like `gsplatZaxis`, which returns a unit vector in that direction. This PR adds this helper and the symmetric `gsplatXaxis`/`gsplatYaxis` as well.

### DIFF
--- a/docs/docs/dyno-stdlib.md
+++ b/docs/docs/dyno-stdlib.md
@@ -232,4 +232,7 @@ This way, you can pass around a `DynoVal<Gsplat>` that contains the all the prop
 | `splitGsplat(gsplat)` | Split a `Gsplat` into its components |
 | `combineGsplat(gsplat)` | Create a `Gsplat` from components (or inject components into an existing `Gsplat`) |
 | `gsplatNormal(gsplat)` | Get the `Gsplat` normal, defined as whichever X/Y/Z axis has the smallest scale |
+| `gsplatXAxis(gsplat)` | Return the unit vector along the splat's X-axis |
+| `gsplatYAxis(gsplat)` | Return the unit vector along the splat's Y-axis |
+| `gsplatZAxis(gsplat)` | Return the unit Vector along the splat's Z-axis |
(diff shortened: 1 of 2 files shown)


oscarlorentzon:
Looks good in general but I have two questions because these will be in the public API:

- Naming: do we want `XAxis` here rather than `Xaxis`, so `gsplatXAxis`/`GsplatXAxis`?
  - main capitalizes the word after a single letter: `layerYStart`/`layerYEnd` at [`src/SplatAccumulator.ts:416`](https://github.com/sparkjsdev/spark/blob/99e3cede27987bfdd460c8ec58c8bcf8ef4f6db0/src/SplatAccumulator.ts#L416), [`:417`](https://github.com/sparkjsdev/spark/blob/99e3cede27987bfdd460c8ec58c8bcf8ef4f6db0/src/SplatAccumulator.ts#L417), `ProjectHOutput`/`projectHOutputType` at [`src/dyno/vecmat.ts:177`](https://github.com/sparkjsdev/spark/blob/99e3cede27987bfdd460c8ec58c8bcf8ef4f6db0/src/dyno/vecmat.ts#L177), [`:183`](https://github.com/sparkjsdev/spark/blob/99e3cede27987bfdd460c8ec58c8bcf8ef4f6db0/src/dyno/vecmat.ts#L183). three.js uses `xAxis`/`yAxis`/`zAxis` in `Matrix4.makeBasis` and `extractBasis`.
- Docs: do we want rows for these in the dyno stdlib table, next to `gsplatNormal` at [`docs/docs/dyno-stdlib.md:234`](https://github.com/sparkjsdev/spark/blob/da2fe65736f3bf867342aaab838988d5ebee753b/docs/docs/dyno-stdlib.md#L234)?

asundqui:
Updated!

oscarlorentzon (approved):

PR #420, opened 2026-09-03
Avoid LoD texture upload stalls during .rad animation

### LINKED ISSUE #347: `.rad` files causing stutter during animation
Hi, thank you for all the work you've put into Spark! I'm working on animating splats (all `.rad` files) by parenting them to a GLTF and animating it. However, I'm noticing some stutter when running an animation for the first few runs:

I switched out the splat files for `.ply`s and a `.spz`, and the animation runs smoothly. I've experimented by only having one `.rad` and the rest be other types, and the stutter gets reintroduced even with one `.rad` file.

miakh:
## Summary

- replace sampled LoD index textures instead of updating them in place
- upload replacement ordering textures in four-row chunks and swap them in only after the upload completes
- retain one previous texture generation so GPU work that still references it can finish safely
- add focused coverage for texture replacement/disposal and row-offset uploads

## Why

Updating a texture that the preceding frame is still sampling can synchronously wait for GPU work. With animated `.rad` splats, repeated LoD updates turn those waits into visible frame-time spikes during the first animation runs.

This keeps the existing LoD and sorting contracts intact. It changes only the texture upload lifecycle: build an unreferenced replacement, upload it incrementally, then atomically expose it to rendering.

## Performance evidence

Reporter-derived reproduction: the same 2.5M-splat `.rad` scene, viewport, camera orbit, and 1.25 s GLTF animation on both revisions.

| Metric | Before | After |
| --- | ---: | ---: |
| Average frame time | 19.2 ms | 16.7 ms |
| 95th percentile | 67.7 ms | 18.7 ms |
| Worst frame | 93.4 ms | 32.6 ms |
| Frames over 50 ms | 4 | 0 |

## Video proof

The same orbit and animation are shown before and after; watch the frame-time timeline and the first-run stalls disappear in the fixed build.

https://github.com/user-attachments/assets/bc2ce366-403f-4865-b792-99d40ed143f0

## Validation

- `npm run lint:summary --silent`
- `npm test --silent` (3/3 passing)
- `npm run build:dev --silent`
- reporter-derived browser reproduction with no page errors

Fixes #347

### DIFF
--- a/src/SparkRenderer.ts
+++ b/src/SparkRenderer.ts
@@ -11,6 +11,10 @@ import {
 import { SplatAccumulator } from "./SplatAccumulator";
 import { SplatGeometry } from "./SplatGeometry";
 import { SplatWorker } from "./SplatWorker";
+import {
+  disposeU32DataTextures,
+  replaceU32DataTexture,
+} from "./U32DataTexture";
 import { SPLAT_TEX_HEIGHT, SPLAT_TEX_WIDTH } from "./defines";
 import { getShaders } from "./shaders";
 import {
@@ -21,6 +25,8 @@ import {
   isVisionPro,
   uploadU32DataTextureRows,
 } from "./utils";
+
+const ORDERING_UPLOAD_ROWS_PER_FRAME = 4;
 
 export interface SparkRendererOptions {
   /**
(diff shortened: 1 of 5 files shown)


oscarlorentzon:
Thanks for this, and welcome.

The diff does three things at once:

1. It builds a new texture on each update and swaps it in, instead of writing into the old one.
2. It keeps the previous texture around for one more update.
3. It uploads the new ordering texture 4 rows per frame instead of all at once.

I have a few questions on the measurements:

- What do 1 and 2 do to the upload time itself?
- Were the three measured separately, or only together?
- How long does 3 hold the sort result, and how does that scale with splat count?

One more question, on the delay. Does the sort order lag visibly during fast camera moves, and does that get worse with more splats?

Also, from the console. On a MacBook in Chrome I get these:

```
WebGL: INVALID_VALUE: texSubImage2D: no pixels     SparkRenderer.ts:1119
```

That line is the `initTexture` call on the new texture.

miakh:
Thanks for pointing this out. After futher testing, and taking full scene in account, I cannot see that it makes it faster.  The chunked upload also delays sorting more as splat count grows, so I’m closing this PR.

PR #424, opened 2026-09-11
Allocate all four SH textures for paged ext splats

oscarlorentzon:
Loading an asset with SH degree 3 throws `RangeError: offset is out of bounds` when both `paged` and `pagedExtSplats` are on. It happens once for each page that has SH3 data in it. The page is left without one of its two SH3 textures, and the rest of that frame's loading is skipped, so the scene fills in slowly or not at all.

* With `pagedExtSplats`, SH3 is stored in two textures instead of one, so an sh3 asset [needs four SH textures](https://github.com/sparkjsdev/spark/blob/728fc4053904e03c6f63f524a2ea4e4793935cbe/src/SplatPager.ts#L842) where other assets need three. [Only three were created](https://github.com/sparkjsdev/spark/blob/728fc4053904e03c6f63f524a2ea4e4793935cbe/src/SplatPager.ts#L999-L1001), so the fourth upload had no texture to write into and failed.
* The fix creates as many textures as the asset has SH data for. Assets with a lower SH degree, and assets loaded without `pagedExtSplats`, work as before.


asundqui:
Nice find and fix! In the future we'll need to add some regression tests so this doesn't happen again.

I'm wondering about removing `this.curSh` though - although it's not used anywhere else in Spark, I'm wondering if we should keep it, so that users can tell what Sh degree is being used? Obviously the logic will have to be different so it doesn't just key off `shArrays.length`, which will be different depending on extSplats vs. not.

oscarlorentzon:
Agree on keeping `curSh` to avoid a breaking change. The same expression is set in `uploadPage` now, since it tracks the SH degree and the allocation tracks textures.

asundqui (approved):
LGTM! cc: @mrxz @dmarcos

mrxz (approved):

dmarcos:
thanks

PR #425, opened 2026-09-14
Sync the lock file with the manifest

oscarlorentzon:
This updates the out-of-date lock file version so `npm install` stops changing it every time. The license line changes too after `npm run build:wasm` has been run.


asundqui (approved):
Yes, agreed we should update the lock file. cc: @dmarcos

dmarcos:
thanks

PR #426, opened 2026-09-14
Point repository metadata at the correct GitHub organization

oscarlorentzon:
The repository URL had an extra hyphen, so the link shown on the published npm package page, in the Rust crate metadata and in the examples page header returned a 404.

Correct the organization name in the package manifest, the Rust workspace manifest and the examples page so every link resolves to the real repository.


asundqui (approved):
Seems like a good idea yes! @dmarcos

dmarcos:
thanks

PR #427, opened 2026-09-14
Add a typecheck script and run it in CI

oscarlorentzon:
Start checking types to avoid type errors. Add a script that checks types without building and run it in Linux CI workflow. 

The TypeScript 6 upgrade fixes an existing type error with no code change, and the generated type files stay the same so it should not affect compatibility.


oscarlorentzon:
The first version had no cache, and building the wasm package took close to two minutes.
The latest update adds a cache for the Rust build.

asundqui (approved):
Love it, thanks for this improvement.

oscarlorentzon:
Rebased to fix merge conflicts with #445.

dmarcos:
thanks!

PR #430, opened 2026-09-16
Fix RangeError uploading degree-3 SH to paged extended splats

xohm:
### The bug

With `pagedExtSplats: true`, SH degree 3 is stored in two textures (SH3A and SH3B), so a degree-3 chunk arrives as **four** SH arrays. `SplatPager.uploadPage` sizes the textures with:

```ts
const numSh = Math.min(shArrays.length, 3);
this.ensureShTextures(numSh);
```

`numSh` is an SH **degree**, but `ensureShTextures` spends it as a texture **count** (`for (let i = 0; i < this.curSh; i++)`). Only three textures are allocated; the fourth stays the 8192-element placeholder, and the first page written into it throws:

```
RangeError: offset is out of bounds
  at Uint32Array.set
  at uploadTextureLayer
  at SplatPager.uploadPage
  at SplatPager.processUploads
  at SparkRenderer.updateLodInstances
```

### Why it matters beyond the error

The throw escapes `processUploads` and leaves `updateLodInstances` before it sets `pager.fetchPriority`, so the pager never asks for another chunk: the scene stops at its first one and stays at its coarsest LOD, which reads as slow streaming rather than a crash. `SH3B` is also never filled, so the third SH band is wrong wherever it does draw.

Measured on a 6.4M-splat `.rad` with `maxSh: 3` (`lodSplatCount` 300k, headless Chromium/SwiftShader, 90 s):

| | resident pages | page errors |
| --- | --- | --- |
| before | 1 | 1 |
| after | 70 | 0 |

### The fix

Keep `curSh` a degree and allocate one more texture when extended pages reach degree 3.

### Notes

- Reachable since #404. Before that, the same input threw `Invalid array length` earlier in `processFetched`, so the four-array list never got as far as `uploadPage`.
- Needs paged + `pagedExtSplats` + `maxSh: 3` together. The `streaming-lod` example does not show it because those scenes are SH degree 0 and 1.
- Source only; `dist/` is left to your build step.


xohm:
Closing for now — opened before our internal review was finished. We will resubmit once it is done.

oscarlorentzon:
@xohm Thanks for the contribution. This should be when #424 is merged.

xohm:
Thanks — you're right, #424 is the same fix and it predates this one; I should have checked the open PRs before writing it. Its form is better too: making `ensureShTextures` take a texture count removes the degree/count confusion rather than working around it. We've taken #424 into our build in the meantime.

In case it's useful for #424: the throw also leaves `updateLodInstances` before it sets `pager.fetchPriority`, so the pager stops requesting chunks and the scene stays at its coarsest LOD. On a 6.4M-splat `.rad` with `maxSh: 3` (`lodSplatCount` 300k, headless Chromium/SwiftShader, 90 s) we measured 1 resident page before and 70 after.

PR #431, opened 2026-09-16
Add a fetchBytes hook to PagedSplats

xohm:
### Why

`PagedSplats` can page a RAD that is a plain ranged URL (`rootUrl`), or one already whole in memory (`fileBytes`). Neither fits a RAD that is:

- stored **inside a container** at an offset (our case: a `.rad` member of a project archive),
- kept in **OPFS or IndexedDB**,
- served behind **bespoke authentication** (signed requests, custom headers per range),
- **picked by the user** as a local `File`, which has no URL at all — and on a page opened from `file://` there is no server to range against either.

`fetchRange` is module-private, so none of this can be supplied from consumer code today.

### What

```ts
export type FetchBytes = (
  offset?: number,
  bytes?: number,
  signal?: AbortSignal,
) => Promise<Uint8Array>;

new PagedSplats({ fetchBytes, fileType: SplatFileType.RAD });
```

When set, `getRadMeta` and `fetchDecodeChunk` ask the hook for their ranges instead of calling `fetchRange`. Returning **fewer** bytes than asked for is allowed and means the file ended, which lets the header probe stop backing off on a small file. The `PagedSplats`' own abort signal is passed to the hook, so a read in flight is cancelled on dispose the way the URL path's fetch is.

RAD only, and `fileType` must be given, since there are no bytes to sniff before the first read. Sibling-file chunked RAD and the existing `rootUrl` / `fileBytes` paths are untouched — nothing changes for anyone who does not pass the hook.

### Verified

Built from this branch and driven in headless Chromium (SwiftShader) against a 106 MB archive holding a `.rad` member: the hook is used for the header and for chunk reads, the header decodes, the first chunk decodes to 48,196 splats, and opening the file costs 18 KB of reads. The same hook backs a `file://` page with no server, reading through `File.slice`.

We have been running this in production for two weeks in our own viewer, over both local files and containers.


xohm:
Closing for now — opened before our internal review was finished. We will resubmit once it is done.

PR #434, opened 2026-09-16
Rotate around the controlled object's up axis

### LINKED ISSUE #300: PointerControls assumes Y-up world coordinate system internally, causing rotation to behave incorrectly when camera.up is Z-up.
PointerControls assumes Y-up world coordinate system internally, causing rotation to behave incorrectly when camera.up is Z-up.

oscarlorentzon:
Closes #300.

Dragging, the arrow keys and the gamepad all turned around Y, so with any other up
axis the view tilted as it turned and the up and down limit stopped the sideways
turn at 90 degrees.

Turning now follows the up axis of the object passed in, which defaults to Y.


asundqui (approved):
Great! Let's merge this in, this is how we should have been doing it.

dmarcos:
Thanks!

PR #435, opened 2026-09-16
Fix pinch and press movement under a rotated parent

oscarlorentzon:
Pinch and press gestures aimed the raycaster with a direction measured against the world, then applied it to a
position measured against the object that is the parent of the camera, so a rotated parent
sent the ray in the wrong direction.


asundqui (approved):
Great! I'm assuming you've tested these scenarios and it works as expected :)

oscarlorentzon:
Yes, tested both gestures manually and they show equivalent faulty direction on main and move straight forward correctly after the fix in the rotated parent case.

**Baseline**

https://github.com/user-attachments/assets/74b9e5ef-e156-4299-8c48-d92fcc083e4f

**Fix**

https://github.com/user-attachments/assets/14f99ac4-a1de-455f-8846-b61abdb5ce52

dmarcos:
thanks

PR #436, opened 2026-09-16
Skip pointer frames that took no time

oscarlorentzon:
Inertia divides distance by frame time, so a frame of no time divides by zero and
breaks the camera. Such frames are now skipped, and their potential movement executes
on the next frame.


asundqui (approved):
Great!

dmarcos:
Thanks!

PR #438, opened 2026-09-17
Await wasm initialization and check sh degress in transcodeSpz

### LINKED ISSUE #437: Calling transcodeSpz before WASM initialization completes fails in v2.2 (works in v2.1)
Agent-produced description of the problem:

Both happen inside Spark’s transcodeSpz, which converts a Gaussian-splat file into compressed SPZ.

1. The API runs before WASM is ready.
Importing Spark starts loading its WebAssembly engine asynchronously. The import finishes before that initialization does, but transcodeSpz immediately calls the engine. Its internal wasm object is still undefined, so accessing __wbindgen_malloc—the WASM memory allocator—throws.
Upstream fix: make transcodeSpz await Spark’s initialization promise before calling WASM. Callers shouldn’t need a delay.

2. The transform path crashes on an ordinary splat.
After waiting, WASM is ready and Spark decodes the PLY. It then calls GsplatArray.transform, even though we requested no transformation. Our input has one Gaussian and no level-of-detail hierarchy. That path attempts to access index 0 of an empty array and panics. JavaScript exposes the Rust panic as RuntimeError: unreachable.
Upstream fix: make the transform implementation handle non-LoD inputs correctly. Skipping the unnecessary transform avoids this reproducer’s crash, but doesn’t fix that underlying implementation.

mrxz:
Fixes #437 

This PR intends to resolve the two issues described in #437. Although I wasn't able to reproduce the first issue, the `transcodeSpz` method does indeed not await initialization of the WASM module before calling several of its exported methods. Introducing an await should resolve it.

As for the second issue, it isn't tied to LoD vs non-LoD inputs, but instead a missing check on the sh degrees of the inputs. The fact that the transform is always executed is simply inherited from the original JS based implementation. While it might be worthwhile to skip this when not needed, that would not resolve the underlying issue.


oscarlorentzon:
The fix looks good. This should be ready to merge @dmarcos.

I can reproduce the first issue in node with:

```ts
import { transcodeSpz } from "@sparkjsdev/spark";

await transcodeSpz({
  inputs: [{ fileBytes: new Uint8Array(), fileType: "ply" }],
});
```

oscarlorentzon (approved):

dmarcos:
thanks

PR #439, opened 2026-09-18
Use the default encoding when none is given

oscarlorentzon:
A `PackedSplats` has an encoding only when one is passed to it, and `SplatLoader` passes one when it reads a file. Writing a `PackedSplats` with no encoding to SPZ failed, because two of the conversions that take an encoding treated a missing encoding as an error, where the others fall back to the default. Those two now fall back as well. #415 fixes the same failure from the JavaScript side.


asundqui (approved):
That's the right behavior, thanks!

dmarcos:
Thanks!

PR #440, opened 2026-09-18
Accept extended splats when writing SPZ

oscarlorentzon:
`writeSpz` accepted only `PackedSplats`, so an `ExtSplats` could not be written. `ExtSplats` is now accepted too, and its positions reach the encoder as the 32 bit floats it stores rather than the half floats `PackedSplats` holds. Calls that pass a `PackedSplats` still work. The types describing what `transcodeSpz` accepts are now exported, so an application can import them instead of restating the shape.


asundqui (approved):
So fast! Thank you.

dmarcos:
Thanks!

dmarcos:
needs rebase

oscarlorentzon:
Rebased.

PR #441, opened 2026-09-18
Restore version 3 as the default SPZ output

oscarlorentzon:
Releases from 0.1.9 to 2.1.0 wrote SPZ version 3 and 2.2.0 switched to version 2. Version 3 is the default again, with version 2 selectable through a new options argument, and existing calls still work. 

The version 3 quaternion packing is fixed too, so a rotation encodes and decodes unchanged. Reading is unaffected and still accepts versions 1 to 3.


asundqui (approved):
LGTM. Thank you for adding the test as well!

dmarcos:
thanks!

PR #442, opened 2026-09-18
Scale the gamepad stick by the rotate speed once

oscarlorentzon:
`FpsMovement` scaled the rotation stick by `rotateSpeed` twice where the rotate keys were scaled once, so the stick turned at `rotateSpeed` squared. The stick now turns at `rotateSpeed`, like the keys. The XR controller axes are added into the same stick, so XR rotation was scaled twice too. Roll turns at `rollSpeed` as before, since the stick sets only yaw and pitch.


asundqui (approved):
suchSpeed^2!

dmarcos:
Thanks!

PR #445, opened 2026-09-21
Update TypeScript and Vite, use Vitest for tests

mrxz:
This PR updates several of the `devDependencies`, mainly motivated by the goal of updating TypeScript, where the other changes followed naturally:

- TypeScript updated to version 7.0.2
- Newer version of Vite required for TypeScript 7
- Updated version of `vite-plugin-glsl` needed for Vite 7+ support
- `vite-plugin-dts` can be made to work, but effectively requires including TS 6.X as well, instead replaced with native tsc
- `ts-node` does not work with TypeScript 7.0.2 (https://github.com/TypeStrong/ts-node/issues/2174), replaced with Vitest for unit tests

Removing `vite-plugin-dts` has several benefits, as it turns out this was a good chunk of the build time (~90%), additionally it was performed for both dev and production builds. The output of `tsc` is near identical, only differing in things like used quotes and order of properties/types (see https://github.com/mrxz/spark/commit/53835458d3d917edac45a45f49d0f9b3d3e6e226).

Adding Vitest to replace the `ts-node` based "test runner" is up for debate. I do think a proper testing framework makes sense over raw node asserts, but it doesn't have to be Vitest. Given Vite is already used though, it is a natural choice IMHO.


tsconfig.json
oscarlorentzon: Can this block be removed now that the `ts-node` dependency is gone?

oscarlorentzon (approved):
Looks good overall, and Vitest is a reasonable choice with Vite already in use.

tsconfig.json
mrxz: Good catch, this can indeed be removed now.

dmarcos:
Thanks!

PR #446, opened 2026-09-21
Identify Three.js types with instanceof

oscarlorentzon:
Replaces #443.

Three.js types were identified two ways, `instanceof` in 20 places and the `isTexture` marker in two. The library now uses `instanceof` throughout, apart from the XR render target check, where a comment says why.


mrxz (approved):

dmarcos:
Sweet! Thanks!

PR #447, opened 2026-09-21
Read and apply `maxSh` in `PagedSplats`

oscarlorentzon:
`PagedSplatsOptions` declares `maxSh`, but the `PagedSplats` constructor took the limit from the `SplatPager` it was given and never read the option. The constructor now reads the option, with the same default of 3, and the pager limit still bounds the SH level, so a `PagedSplats` given a pager and no option renders as before.

`PagedSplats.hasRgbDir` tested the SH level in the file against the `SplatPager` limit alone, though the SH count passed to the shader already takes both limits. The test now takes the limit set on the source too, as `PackedSplats` and `ExtSplats` do, so a `SplatMesh` with `maxSh` set to 0 drops the per-splat view direction and SH evaluation from its generator and stops marking the mesh for regeneration when the camera moves. Rendered color is unchanged.


dmarcos:
Thanks!

PR #448, opened 2026-09-21
Wait for WebAssembly setup before writing SPZ

oscarlorentzon:
`writeSpz` called into the WebAssembly module without awaiting the instantiation that starts when Spark is imported, so a call made before instantiation finished threw `TypeError: Cannot read properties of undefined (reading 'packedsplats_to_gsplatarray')`. Awaiting instantiation makes `writeSpz` async, so both overloads return a promise and every caller has to await the result, a break in the signature added in 2.2.0. 

`transcodeSpz` added an equivalent await in #438.


oscarlorentzon:
As mentioned above, this breaks the `writeSpz` API introduced in #374. 2.2.0 is the only release with the synchronous signature, and that signature works for any call made after the WebAssembly module has instantiated.

Is it acceptable to break compatibility here if we treat the change as fixing a defect?

mrxz:
> Is it acceptable to break compatibility here if we treat the change as fixing a defect?

IMHO it's acceptable in this case. I expect usage to be low given it was only introduced in 2.2.0, and odds are IO code where it is used is already async and/or can easily be made async. Of course this should be mentioned in the changelog/release notes.

An alternative could be to introduce a new method that awaits wasm initialization and deprecate the current `writeSpz` method (JSDoc annotation + logging deprecation warning). Though we don't really have a deprecation policy yet, and it's likely more work than its worth for this.

mrxz (approved):

dmarcos:
needs rebase

oscarlorentzon:
Rebased.

asundqui:
@dylanhu7 FYI I've seen we use this function in a few places, but it looks to me like making them async should work for all of them.

PR #450, opened 2026-09-22
Cancel scheduled work when disposing the renderer

oscarlorentzon:
`SparkRenderer.dispose()` left the update and sort timeouts pending. The update callback then ran on the disposed renderer and allocated a `WebGLArrayRenderTarget` and a LoD `SplatWorker`. `dispose()` now clears both ids.


asundqui (approved):
Excellent, good catch.

dmarcos:
Thanks!

PR #451, opened 2026-09-23
Fix paged root eviction issue

asundqui:
This PR builds on top of #449  and fixes the issue labeled P2 from #428 . Only the last commit in this PR is stacked on top of #449. Once that's merged I will rebase this on main. The delta can be seen here: https://github.com/asundqui/spark/compare/fix-paged-lod-browser-test...asundqui:spark:fix-paged-root-eviction

When the splat page table is full, if a paged splat is hidden or removed and its root (page 0) is evicted, we fail to clear its internal `rootPage`. As a result, if another splat object loads new data into that same page and the old splat becomes visible again within the 3 sec LoD disposal timeout, it will traverse from that same page and show incorrect data. The fix is simply to clear the rootPage and the displayed splats for that object if it's evicted:
```
           } else if (!lodTree && chunk === 0 && record.rootPage === page) {
              // Root chunk evicted: forget the page (traversal skips this mesh
              // until the root is refetched) and stop drawing indices into
              // pages that no longer hold this mesh's data.
              record.rootPage = undefined;
              splats.clear();
            }
```

This PR also adds a new test case to `paged-lod.test.ts` that fails before this fix but now succeeds. It creates a page table with a capacity of 1, adds two SplatMeshes where only one is visible. Then it switches visibility to the other splat, and sets `enableLodFetching = false` which evicts the previous splat page but prevents the new one from loading. This creates a scenario where nothing should be rendered, but without this fix we see the root splat from the pre-existing data rendered.

In order for this test to run deterministically we also add an option `SparkRenderer.lodCleanupTimeoutMs` which previously was hardcoded to 3s. Setting it to Infinity allows the above test to detect the failure even if it takes longer than 3s for the scenario to settle.

Finally, in addition to providing an adjustable `lodCleanupTimeoutMs` this PR will dispose all LoD objects that exceed the timeout, rather than only one per LoD work cycle.


test/browser/paged-lod.test.ts
```diff
@@ -25,3 +25,68 @@ for (const [lodSplatCount, snapshot] of [
...
+// left recovers to its original render.
+test("re-shows a paged mesh whose root page was evicted", async ({
+  harnessPage,
+}) => {
+  const { leftAlone, evicted, recovered } = await harnessPage.evaluate(
+    async () => {
+      const h = window.harness;
+      const spark = h.createSpark({
+        lodSplatCount: 10_000,
+        maxPagedSplats: 65536,
```
mrxz, line 43: Nitpick: would be nice to make abundantly clear that this "magic number" is exactly one page. Don't think we export any constant for it, so something like `65536 /* 1 page */` is fine IMHO

mrxz (approved):

src/SparkRenderer.ts
```diff
@@ -1541,39 +1557,44 @@ export class SparkRenderer extends THREE.Mesh {
...
-        oldest = {
-          splats,
-          lastTouched: record.lastTouched,
-          lodId: record.lodId,
-        };
+    // Dispose every expired tree, one at a time. The oldest record is re-evaluated
+    // after each await: lodIds is only added to/removed from inside this exclusive
+    // callback, but driveLod() may bump lastTouched between awaits, and a tree that
+    // was touched in the meantime must not be disposed.
+    while (true) {
```
oscarlorentzon: Something to consider for a follow-up: this could collect every expired tree in one synchronous pass and dispose them all with a single `disposeLodTree` call, so there is no re-scan after each `await`.

oscarlorentzon (approved):
Looks good overall 👍 .

test/browser/paged-lod.test.ts
asundqui: Added the clarifying comment as suggested.

src/SparkRenderer.ts
asundqui: Good idea. At first I thought it would be unsafe due to awaiting each `disposeLodTree` call, but we can make all changes to `lodIds*` and `lodMeshes` in a synchronous step in one go. I kept a loop that disposes one lodTree at a time rather than disposing in a batch because I wanted to keep this PR minimal and not modify Rust code etc to allow batch `disposeLodTree`. The latest commit also fixes a potential issue when `lodCleanupTimeoutMs=0`, by excluding any actively rendered `lodMeshes`.

asundqui:
Addressed feedback, rebased, all tests including browser tests successful.

dmarcos:
needs rebase after landing #449

PR #452, opened 2026-09-23
Fix paged race condition with late chunk arrival. Add spark hooks testing infra

asundqui:
This PR builds on top of #451 and #449 and fixes issues labeled P1a+P1b from earlier PR #428 . Only the last commit in this PR is stacked on top. Once the two previous PRs are merged I will rebase this onto main before merging. The delta can be seen here: https://github.com/asundqui/spark/compare/fix-paged-root-eviction...fix-paged-late-chunks-add-spark-hooks

Two new browsers tests are added, which deterministically fail before this PR and now succeed:
1. `paged-lod.test.ts: drops a chunk that lands after its LoD tree was cleaned up`: If a paged mesh is hidden and its LoD tree cleaned up while a chunk fetch is still in flight, the chunk landing afterwards was "leaked". SplatPager recorded it as resident but SparkRenderer had no tree for it, so when the mesh was shown again the pager never re-fetched the root and the mesh rendered nothing, forever.
2. `paged-lod.test.ts: purges queued updates when a LoD tree is cleaned up mid-callback`: Similar, but the chunk lands just before the tree is cleaned up, after Spark has already consumed pager updates for that frame. Its insert and upload stayed queued while removeSplats freed the page, so when the mesh was shown again the stale insert was applied to the new tree and the mesh rendered from a page the pager considered free.

In order to make these fixes easy to test, this PR has a temporary bool flag in `SplatPager.ts`: `FIX_LATE_CHUNKS`. Toggling it from true to false makes the two new tests fail. Before merging this I will remove the flag (trivial edit). The fix is to check if `SparkRenderer` still considers the object active (via callback `isActive` and `SparkRenderer.lodIds`) when chunks arrive, and to remove any queued/pending chunks in `SplatPager.removeSplats` which is called during `SparkRenderer.cleanupLodTrees`.

In order to create deterministic tests for these scenarios, this PR borrows ideas from #428 by adding a "Spark hooks" testing infrastructure. Unlike the earlier PR this is minimal and much simpler/cleaner, and is also more expressive. The hooks are gated behind an environment variable `SPARK_ENABLE_HOOKS=1` which is only enabled during browser testing and is removed during normal builds. We also modify the browser tests so they run in a separate Vite dev server so as not to conflict with one you may already have running.

Along with the env variable, hooks are enabled by calling `setSparkHook(...)` with a function that takes `(name, context?)` and returns either `undefined` or a Promise. This allows us to create async await points that can use both the name and arbitrary context to decide whether to hold this hook point. For test #2 above we add a single hook point `lod.beforeCleanup` just before `SparkRenderer.cleanupLodTrees` which allows us to receive the chunk at the desired point.

The browser testing harness adds a method `holdHook(name, match?)` which enables a hook point by name and an optional callback to check the context and conditionally match. The method returns `{ reached, release }` which allows the tester to await reaching that hook, and a function it can call to release the hold.

We also add a harness method `holdRequest(url)` that allows us to programmatically pause globbed URLs, and returns `{ requested, release }` which similarly allows us to wait until such a URL is actually requested and subsequently release it when we're ready. This is used to delay chunk fetches in our new tests until we want them to arrive.

Finally, we amend the harness method `settle()` to allow it to ignore pending fetches, since our test setup needs everything else to settle while we hold back the fetch.


asundqui:
Successful browser test run via GitHub CI for this commit: https://github.com/asundqui/spark/actions/runs/35911625832

asundqui:
Claude found an additional possibility where if `SplatPager.removeSplats` ran between the `.then` and `.finally` when fetching chunks and we didn't yet have any pages mapped, the early return wouldn't run this fix to clear out the obsolete chunk. Not adding another test for this as it would require another hook point in the `.finally` block.  Updated browser test results: https://github.com/asundqui/spark/actions/runs/35930346580

src/hooks.ts
```diff
@@ -0,0 +1,29 @@
...
+export type SparkHook = (
+  name: string,
+  context?: Record<string, unknown>,
+) => void | Promise<void>;
+
+// import.meta.env only exists under Vite/Vitest; plain Node has no env object.
+const env = (import.meta as { env?: Record<string, unknown> }).env;
+
+/** True only when built with SPARK_ENABLE_HOOKS=1. Gates every hook point. */
+export const SPARK_ENABLE_HOOKS: boolean = env?.SPARK_ENABLE_HOOKS === "1";
```
oscarlorentzon: Through `env`, it seems like the build inlines the whole `import.meta.env` object (and any `SPARK_*` variable set during the build), so the flag is checked at runtime and the hooks code is in every `dist` bundle. With the direct read it is `false` at build time and the hooks code is removed. The browser tests still get `true` from `SPARK_ENABLE_HOOKS=1` in `playwright.config.ts`.

```suggestion
/** True only when built with SPARK_ENABLE_HOOKS=1. Gates every hook point. */
export const SPARK_ENABLE_HOOKS: boolean =
  import.meta.env?.SPARK_ENABLE_HOOKS === "1";
```

src/index.ts
```diff
@@ -111,6 +111,8 @@ export { LN_SCALE_MIN, LN_SCALE_MAX, SplatFileType } from "./defines";
 
 export * as defines from "./defines";
 
+export { SPARK_ENABLE_HOOKS, setSparkHook, type SparkHook } from "./hooks";
```
oscarlorentzon: Do we want these in the public API? Only the test harness uses them, and it can import them from `hooks.ts` instead:

```ts
import { SPARK_ENABLE_HOOKS, setSparkHook } from "../../src/hooks.js";
```

oscarlorentzon (approved):
Looks good in general. Two small questions inline.

src/index.ts
asundqui: Agreed, removing it from the public API.

src/hooks.ts
asundqui: You're right - I had seen this issue in the previous mega-PR but failed to do it correctly here.

I made the change you suggested, but apparently the optional `env?` isn't necessary - it works without it. Even more interestingly, with your other suggestion (removing these exports from `index.ts`) - Rollup is apparently able to figure out that since `setSparkHook` isn't exported, it can never be called, hence `hook` is always undefined, and the hook points never do anything, so it gets removed automagically that way anyway!

asundqui:
All tests including new browser tests re-run and successful.

asundqui:
@dmarcos this PR still has a temporary "for easy reviewer testing" flag `FIX_LATE_CHUNKS`. Was curious to hear @mrxz 's feedback on this PR first. I'll remove the flag (trivial edit) before any merging (but no harm either if it's merged with the flag, easy to remove later).

src/utils.ts
```diff
@@ -328,6 +328,13 @@ export function newArray<T>(
   return new Array(n).fill(null).map((_, i) => initFunction(i));
 }
 
+// Remove every element matching predicate from the array, in place
+export function removeWhere<T>(array: T[], predicate: (item: T) => boolean) {
+  for (let i = array.length - 1; i >= 0; i--) {
+    if (predicate(array[i])) array.splice(i, 1);
+  }
```
mrxz: Repeated `array.slice` can be very inefficient, especially for larger arrays. For the arrays it's currently used on that likely isn't a big issue, but with how generic this function is in `utils` (which is also exported) that might make it less ideal.

_Iff_ the order of the array elements doesn't matter, the to be deleted elements can be swapped with the tail/end of the array. This allows the arrays length to be shrunk and keeps the number of elements moved O(n).

src/SplatPager.ts
```diff
@@ -1088,10 +1114,14 @@ export class SplatPager {
           .fetchDecodeChunk(chunk)
           .then(
             async (data) => {
-              // Make sure the originating PagedSplat hasn't been disposed in the meantime
+              // Drop the chunk if the originating PagedSplats was disposed, or
+              // its LoD tree was cleaned up, while the fetch was in flight.
               if (splats.abortController.signal.aborted) {
                 return;
               }
+              if (FIX_LATE_CHUNKS && !this.isActive?.(splats)) {
```
mrxz: Frankly the `isActive` callback to get information from the `SparkRenderer` feels a bit makeshift. Ideally the `SplatPager` should be able to know by itself whether this request is still needed.

Would it be an option to check if the current `fetcher` is still part of `this.fetchers` instead?

test/browser/playwright.config.ts
```diff
@@ -1,8 +1,25 @@
...
+  return new Promise<number>((resolve, reject) => {
+    const server = net.createServer();
+    server.once("error", reject);
+    server.listen(0, () => {
+      const { port } = server.address() as net.AddressInfo;
+      server.close(() => resolve(port));
+    });
+  });
+}
+process.env.SPARK_TEST_PORT ??= String(await freePort());
```
mrxz: I'm fine with the `freePort` logic, just wondering if we'd really need it? Just picking an alternative port as the default, while still allowing `SPARK_TEST_PORT` to use a specific one also suffices IMO.

mrxz:
Looks pretty good! Really like the test side of `holdHook` and `holdRequest` returning objects, which makes their usage in the tests nice and clear.

I did verify that the tests do indeed cover the described issues by toggling the `FIX_LATE_CHUNKS` variable. Also checked the resulting bundles and Vite/Rolldown is indeed able to eliminate it.

Left some comments, but otherwise I think it's pretty much good to go.

test/browser/playwright.config.ts
asundqui: You're right, it's needlessly complicated. I'll do what you suggest, in the next commit.

src/utils.ts
asundqui: You're right, that's needlessly inefficient. I replaced it with a 1-pass in-place removal. Some of them are used as FIFOs so doing an in-place in-order removal seems handy.

src/SplatPager.ts
asundqui: Yes you're right, using `this.fetchers` is much cleaner, thanks for the suggestion. Will do this change.

asundqui:
@mrxz addressed your comments with the last commit. All tests pass including browser tests.

I think this is ready to go @dmarcos !

PR #453, opened 2026-09-23
Use the calling renderer's encoding for its splats

oscarlorentzon:
A second `SparkRenderer`, such as one rendering another viewpoint to a target, draws its splats through the `SparkRenderer` in the scene.

When the two renderers were constructed with different `accumExtSplats` or `covSplats`, the second renderer's splats were decoded with the scene renderer's encoding. The encoding now comes from the renderer whose splats are drawn. Rendering with only the scene renderer is unchanged.

### Verification

- `examples/multiple-viewpoints` still renders its main view and both target viewpoints.


asundqui (approved):
Good catch, thanks!

PR #454, opened 2026-09-24
Write non-finite SPZ rotations as the identity

oscarlorentzon:
Since #441 made SPZ version 3 the default, `SpzEncoder` panics on a rotation with a NaN or infinite component, because the `partial_cmp(...).unwrap()` that picks the largest component panics on NaN. Such a rotation is now written as the identity. 

No other encoder, including SPZ version 2, panics on non-finite input.


asundqui (approved):
Sensible guard to have, thanks!

